PR #6, opened 2020-11-04
JSON loader

jenrik:


cogitantium (approved):
LGTM, I think we should hold off merging this until parser and logger is merged, such that this can be integrated in a sane way.

Mexican-standoff.json is bit-for-bit equal to the generated mexican standoff made by https://github.com/d702e20/encode-game-structure

MathiasMehl (approved):

jenrik:
Merging since the prerequisite #14 have been merged

PR #7, opened 2020-11-04
Graph printer

jenrik:
Print a graph of based on the output of the `succ` function. Can be render to an image using the [`dot` utility](https://linux.die.net/man/1/dot)


NicEastvillage:
Let's see some output !

NicEastvillage (approved):
Looks good (but I don't know `dot` very well). I assume most terminals can handle characters like ⟪⟫, ◯, and 𝑼. Otherwise we can just change to <<>>, O, U, etc.

jenrik:
Here is an example output

graph

src/atl/formula.rs
```diff
@@ -19,5 +20,76 @@ pub(crate) enum Phi {
...
+                        .iter()
+                        .map(|id| id.to_string())
+                        .collect::<Vec<String>>()
+                        .join(",")
+                        .as_str(),
+                )?;
+                f.write_str("⟫(")?;
+                pre.fmt(f)?;
+                f.write_str(" 𝑼 ")?;
+                pre.fmt(f)?;
```
NicEastvillage, line 70:
```suggestion
                until.fmt(f)?;
```
Your graph revealed a bug lmao

NicEastvillage:
I have looked at the example graph and the code for several minutes now. The graph is not supposed to have a dependency on itself. It should have a dependency on a partial move. Is this a bug?

jenrik:
Here is another formula over the mexican-standoff game structure. I also fixed the missing characters in the render by changing the font to Google Noto

```json
{
  "not": {
    "enforce until": {
      "players": [0, 1],
      "pre": {
        "proposition": 0
      },
      "until": {
        "or": [
          {
            "proposition": 2
          },
          {
            "proposition": 1
          }
        ]
      }
    }
  }
}
```

graph dot

src/atl/formula.rs
jenrik: fixed in 214e2f7c860d00c78cb61a3f2e6eedebce554c81

jenrik:
> I have looked at the example graph and the code for several minutes now. The graph is not supposed to have a dependency on itself. It should have a dependency on a partial move. Is this a bug?

That would be a bug in the ATL dependency graph generation code not in the graph printer since it just print whatever `succ` emits. Never the less we might have a big that needs fixing

NicEastvillage (approved):

jenrik:
The bracket for ENFORCE_UNTIL and DESPITE_UNTIL had accidentally been swapped.

Here is an updated graph with partial moves

```json
{
  "not": {
    "despite until": {
      "players": [0, 1],
      "pre": {
        "proposition": 0
      },
      "until": {
        "or": [
          {
            "proposition": 2
          },
          {
            "proposition": 1
          }
        ]
      }
    }
  }
}
```

graph dot

NicEastvillage:
Ah. Nu er alt som det burde være.

src/edg.rs
```diff
@@ -3,21 +3,30 @@ use crate::common::{Edges, HyperEdge, Message, NegationEdge, VertexAssignment, W
...
 use std::collections::{HashMap, HashSet};
+use std::fmt::Debug;
+#[cfg(feature = "graph-printer")]
+use std::fmt::Display;
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
 use std::thread;
-use std::fmt::Debug;
 
+const WORKER_COUNT: u64 = 4;
```
cogitantium, line 13:
```suggestion

```

Seems like this snuck in - there's no usages of it

cogitantium (changes requested):
Looks good. 
I had to delete Cargo.lock before I was able to compile `package `atty` is specified twice in the lockfile`

jenrik:
`Cargo.lock` is being deleted from the repository so that doesn't matter, see #20 and #21

cogitantium (approved):

PR #8, opened 2020-11-06
Primitive AST for Lazy CGS lang

NicEastvillage:
The language will be heavily inspired by PRISM. There's a short description of it in the worksheet. Many features can be added or tweaked, so this aims to give us a foundation which we can build upon.


NicEastvillage:
[Lav AST til lazy CGS sprog](https://trello.com/c/tTFnXOcY/50-lav-ast-til-lazy-cgs-sprog)

cogitantium (approved):
LGTM

PR #14, opened 2020-11-11
Parser and logger

cogitantium:


src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+        edg::distributed_certain_zero(EmptyGraph {}, 0, num_cpus::get() as u64);
+    }
+}
+
+/// Define and parse command line arguments
+fn parse() -> ArgMatches<'static> {
+    App::new("OnTheFlyATL")
+        .version("0.1.0")
+        .author("d702e20 <d702e20@cs.aau.dk>")
+        .arg(Arg::with_name("proposition")
```
lars4509, line 59: I think you mean 'formula'?

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+            .short("p")
+            .long("prop")
+            .env("PROPOSITION")
+            .help("The proposition to check for"))
```
NicEastvillage, line 63: ATL formula, not a proposition

lars4509 (changes requested):
Overall looks fine, but i think you mean ATL formula, instead of proposition

NicEastvillage (approved):

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+            .short("o")
+            .long("log-level")
+            .env("LOG_LEVEL")
+            .default_value("info")
+            .help("{error, warn, info, debug, trace, off}"))
+        .arg(Arg::with_name("log-path")
+            .short("g")
+            .long("log-path")
+            .env("LOG_PATH")
+            .default_value("model-checker.log")
```
jenrik, line 89: how do we disable log files? Can we make it default to not writing any log files?

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+        .arg(Arg::with_name("json_formula")
+            .short("r")
+            .long("jsonformula")
+            .env("JSON_FORMULA")
+            .help("The json to generate formula from"))
+        .arg(Arg::with_name("log-level")
+            .short("o")
+            .long("log-level")
+            .env("LOG_LEVEL")
+            .default_value("info")
```
jenrik, line 83: the UNIX philosophy is to not print anything unless there is an error or the user requested it

```suggestion
            .default_value("warn")
```

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+        .arg(Arg::with_name("json_model")
+            .short("j")
+            .long("json")
+            .env("INPUT_JSON")
+            .help("The json to generate model from"))
+        .arg(Arg::with_name("json_formula")
+            .short("r")
+            .long("jsonformula")
+            .env("JSON_FORMULA")
+            .help("The json to generate formula from"))
```
jenrik, line 78:
```suggestion
            .help("Path to a JSON formatted formula that will be checked against the model"))
```

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+        .arg(Arg::with_name("input_file")
+            .short("i")
+            .long("input")
+            .env("INPUT_FILE")
+            .help("The input file to generate model from"))
+        .arg(Arg::with_name("json_model")
+            .short("j")
+            .long("json")
+            .env("INPUT_JSON")
+            .help("The json to generate model from"))
```
jenrik, line 73:
```suggestion
            .help("Path to the model in JSON format"))
```

src/main.rs
jenrik:
```suggestion
```

src/main.rs
cogitantium: Yes, of course - fixed in ab78e01

src/main.rs
cogitantium: Yes, of course - fixed in ab78e01

lars4509 (approved):

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+    if let Ok(_handle) = log4rs::init_config(get_log4rs_config(
+        args.value_of("log-path").unwrap(),
+        match args.value_of("log-level").unwrap().to_lowercase().as_str() {
+            "error" => LevelFilter::Error,
+            "warn" => LevelFilter::Warn,
+            "info" => LevelFilter::Info,
+            "debug" => LevelFilter::Debug,
+            "trace" => LevelFilter::Trace,
+            "off" => LevelFilter::Off,
+            _ => LevelFilter::Off,
```
jenrik, line 45: We should probably fail with an informative error message instead of silently accepting wrong input

jenrik (changes requested):
The language of the help messages is inconsistent, and they should be reworded in my opinion.

The Tracing framework might be a better alternative for logging and getting diagnostics, see https://tracing.rs/tracing/

src/main.rs
NicEastvillage: updut

src/main.rs
```diff
@@ -20,5 +30,90 @@ impl edg::ExtendedDependencyGraph<i32> for EmptyGraph {
...
+
+        edg::distributed_certain_zero(EmptyGraph {}, 0, num_cpus::get() as u64);
+    }
+}
+
+/// Define and parse command line arguments
+fn parse() -> ArgMatches<'static> {
+    App::new("OnTheFlyATL")
+        .version("0.1.0")
+        .author("d702e20 <d702e20@cs.aau.dk>")
```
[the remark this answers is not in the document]
cogitantium: Fixed in 34082d9

src/main.rs
cogitantium: Added argument to disable and made it default in d69b7c2

src/main.rs
cogitantium: Agreed, this was default in d502e19 project. Fixed in d69b7c2

src/main.rs
cogitantium: Applied in d69b7c2

src/main.rs
cogitantium: Applied in d69b7c2

src/main.rs
cogitantium: Good idea, applied in d69b7c2

PR #16, opened 2020-11-12
Parsing of simple expressions

NicEastvillage:
Adds parsing for simple expressions with +,-,*,/, and (). Other operators will be handled later.

The PR also contains several tests in `parse.rs`.

The parsing is done using [pom](https://crates.io/crates/pom), a parser combinator library. It uses operator overloading to for combinators, which can make it hard to read, but once you understand the basics it becomes super easy to read.

Example: `space() * number() - space()` means
* parse some spaces/whitecharacters (result is `()`)
* then parse a number and throw away prev result (result is now `i32`)
* then parse some more spaces/whitecharacters and keep the prev result (result is still `i32`)

The pom documentation explains it in detail.


NicEastvillage:
[Parsing af simple expressions](https://trello.com/c/4lRpNmLZ/57-parsing-af-simple-expressions)

lars4509 (approved):
Looks good. I just noted that 'alpha()' do not recognize special chars like 'æøå' and such, and it seems like we do not recognize relational operators like '><='.

NicEastvillage:
> 'alpha()' do not recognize special chars like 'æøå' and such
Correct. That's because those characters are not single bytes (`u8`s). We are gonna have to switch to `char`s later I guess - or live without non-ASCII characters

PR #26, opened 2020-11-16
Add subcommands to parser

cogitantium:


cogitantium:
I'm not sure if the return-type of functions `main` and `model_check` are correct. I believe some of the issues stems from having the main loop under the scope of the logger-handler.
Suggestions are welcome.

src/main.rs
```diff
@@ -95,59 +125,83 @@ fn model_check(args: ArgMatches) -> Result<(), Box<dyn Error>> {
...
+                    .long("output")
+                    .env("OUTPUT")
+                    .help("The path to write output to"),
+            )
+    }
+
+    let mut parser = App::new(PKG_NAME)
+        .version(VERSION)
+        .author(AUTHORS);
+    parser = build_common_arguments(parser);
```
NicEastvillage, line 182: This seems to be unused. Is that correct?

src/main.rs
```diff
@@ -63,27 +63,57 @@ fn main() {
...
+    let game_structure = match args.value_of("model_type") {
+        Some("lcgs") => {
+            // TODO: implement lgcs parser
+            EagerGameStructure {
+                player_count: 0,
+                labeling: vec![],
+                transitions: vec![],
+                moves: vec![],
+            }
+        }
```
NicEastvillage, line 103: You have already checked `model_type` in main. Why do it here again?

src/main.rs
cogitantium: Oops, old testing code.

src/main.rs
cogitantium: Because I was unsure where we'll make the lcgs parsing. Fixed in af880a9

src/main.rs
cogitantium: Removed in af880a9

NicEastvillage (approved):
The structure of main could be better to avoid `exit(1)`, but I will disregard it.

cogitantium:
> The structure of main could be better to avoid `exit(1)`, but I will disregard it.

This was from a suggestion from @jenrik during #18, which I agree with.

NicEastvillage:
We could also just return 🤷

MathiasMehl (approved):

cogitantium:
I have removed the feature flags for graph-printing. I don't think it makes sense to have a parser with a graph subcommand and only maybe have the functionality enabled.

GeneralDenmark (approved):

Issue #11, opened 2020-11-07
Missing termination condition


jenrik:
If all worker queues empty, the `distributed_certain_zero` will come to a stand still and never produce a result. The `TerminationDetection` function needs to be implemented as prescribed in Extended _Dependency Graphs and Efficient Distributed Fixed-Point Computation_

PR #36, opened 2020-11-20
Added distinct queue for hyper edges and negation edges

### LINKED ISSUE #12: PickTask bug: Negation edges can be picked when target is unxplored
In alg 1, line 5, we have a `PickTask` function. The requirements for this function is not implemented:

> the function `PickTask` nondeterministically returns:
> - a hyper-edge from W^i_E , or
> - a message from M^i_R or M^i_A, or
> - a negation edge (v,u) from W^i_N provided that A^i (u) ∈ {0,1,⊥}, or
> - a negation edge (v,u) from W^i_N **if all other workers are idle and v has a minimal distance in all waiting lists and message queues** (i.e. for all (v', x) ∈ (W_E ∪ W_N ∪ M_A ∪ M_R ) it holds that dist(v) ≤ dist(v')).
> - If none of the above is satisfied, the worker waits until either a message is received or a negation edge becomes safe to pick. Notice that in this case, W^i_E will remain empty until a message or negation edge is processed.
> -- Last paragraph page 149

and

> Even though PickTask depends on the global state of the computation to decide
whether a negation edge is safe to pick, the rest of the conditions can be decided
based on the data that is available locally to each worker. Therefore it is not
necessary to synchronise across all workers every time a task should be picked,
**it is only required if the worker wants to pick a negation edge (v,u) where
A i (u) = ?.**
> -- First paragraph page 151

The problem relates to negation edges. I have marked it with bold in the citation above. We can only pick a negation edge (v, u) where the assignment of target u is unknown, if all other workers are idle and v has a minimal distance in all waiting lists and message queues.

### LINKED ISSUE #13: Prioritising of Messages
Right now workers do task in the order they are queued, which is FIFO no matter the type of the task. This can result in the worker processing successor edges, even when the parent configuration is final assigned, but the message is later in the queue.  By prioritizing messages of type `ANSWER`, we can ensure that a worker updates already found assignments before computing the successors.

lars4509:
Removed `select()` and instead added an ElseIf-chain, which exhaust the channels in this order: termination, message, hyper, negation. Fix #13 

The change is closer to fix #12, but the case where all workers is idle is still missing, but  since it is strongly related to #11, it is omitted for now.


lars4509:
Hurtig note: Kom til at lave rollback på alt hvad jeg havde lavet over den sidste uge... men fixet det i dag og sørget for at commit mellem hver changes xD (oof)

src/com.rs
```diff
@@ -52,10 +76,21 @@ impl<V: Hash + Eq + PartialEq + Clone> Broker<V> for ChannelBroker<V> {
 }
 
 type WorkQueue<V> = Receiver<Message<V>>;
```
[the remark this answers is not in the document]
jenrik: Seems like a good idea, it's just called that because it used to also contain hyper-edges and negation-edges

src/edg.rs
```diff
@@ -156,97 +189,167 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
-                                Message::ANSWER {
-                                    vertex,
-                                    assignment,
-                                    weight,
-                                } => self.process_answer(&vertex, assignment, weight),
-                            }
+                                // In case of undecided the edge is only processed after 10 iterations
+                                VertexAssignment::UNDECIDED => {
+                                    if self.counter == 9 {
+                                        self.counter = 0;
```
NicEastvillage, line 299: I think we should reset the counter whenever we do something that is not a negation edge too.

src/edg.rs
```diff
@@ -156,97 +189,167 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                    Err(err) => panic!(
+                        "Receiving from negation edge waiting channel failed with: {}",
+                        err
+                    ),
                 }
-                _ => unreachable!(),
+            } else {
+                let _guard = span!(Level::TRACE, "worker release negation", worker_id = self.id);
+                // if the negation channel is empty, unsafe negation edges are released
+                self.release_negations(max(self.unsafe_edges.len(), 0));
```
NicEastvillage, line 318: Unnecessary max

NicEastvillage (changes requested):
Looks good. I am still not entirely sure how distances are update, but I assume you have tested that it works. I only found the few things.

src/edg.rs
```diff
@@ -156,97 +189,167 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
             }
         }
     }
 
+    /// Releasing the edges from the unsafe queue to the safe negation channel
+    fn release_negations(&mut self, dist: usize) {
+        trace!(distance = dist, "release negation");
+        while self.unsafe_edges.len() >= dist && self.unsafe_edges.len() > 0 {
```
jenrik, line 326:
```suggestion
        while self.unsafe_edges.len() >= dist && !self.unsafe_edges.is_empty() {
```

src/edg.rs
```diff
@@ -156,97 +189,167 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+            } else {
+                let _guard = span!(Level::TRACE, "worker release negation", worker_id = self.id);
+                // if the negation channel is empty, unsafe negation edges are released
+                self.release_negations(max(self.unsafe_edges.len(), 0));
             }
         }
     }
 
+    /// Releasing the edges from the unsafe queue to the safe negation channel
+    fn release_negations(&mut self, dist: usize) {
```
jenrik, line 324: this function could use some clean up

src/edg.rs
```diff
@@ -286,6 +387,7 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
                 self.vertex_owner(vertex),
                 Message::REQUEST {
                     vertex: vertex.clone(),
+                    distance: self.distances.get(vertex).unwrap_or(&0).clone(),
```
jenrik, line 390: Fixes `clippy` warning
```suggestion
                    distance: self.distances.get(vertex).unwrap_or(&0),
```

src/edg.rs
```diff
@@ -350,11 +453,26 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
 
     // Mark `dependency` as a prerequisite for finding the final assignment of `vertex`
     fn add_depend(&mut self, vertex: &V, dependency: Edges<V>) {
+        // Update the distance
+        let default = 0;
+        let tdist = self.distances.get(vertex).unwrap_or(&default).clone();
```
jenrik, line 458: Fixes `clippy` warning
```suggestion
        let tdist = self.distances.get(vertex).unwrap_or(&default);
```

src/edg.rs
```diff
@@ -350,11 +453,26 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+        match dependency.clone() {
+            Edges::NEGATION(edge) => {
+                let sdist = self.distances.get(&edge.source).unwrap_or(&default).clone() + 1;
```
jenrik, line 461: Fixes `clippy` warning
```suggestion
                let sdist = self.distances.get(&edge.source).unwrap_or(&default) + 1;
```

src/edg.rs
```diff
@@ -350,11 +453,26 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                self.distances.insert(vertex.clone(), max(sdist, tdist));
+            }
+            Edges::HYPER(edge) => {
+                let sdist = self.distances.get(&edge.source).unwrap_or(&default).clone();
```
jenrik, line 465: Fixes `clippy` warning
```suggestion
                let sdist = self.distances.get(&edge.source).unwrap_or(&default);
```

src/edg.rs
```diff
@@ -350,11 +453,26 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                self.distances.insert(vertex.clone(), max(sdist, tdist));
+            }
+        }
+
+        // Mark `dependency` as a prerequisite for finding the final assignment of `vertex`
         if let Some(dependencies) = self.depends.get_mut(vertex) {
-            dependencies.insert(dependency);
+            dependencies.insert(dependency.clone());
```
jenrik, line 472: Fixes `clippy` warning
```suggestion
            dependencies.insert(dependency);
```

src/edg.rs
```diff
@@ -350,11 +453,26 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
         } else {
             let mut dependencies = HashSet::new();
-            dependencies.insert(dependency);
+            dependencies.insert(dependency.clone());
```
jenrik, line 475: Fixes `clippy` warning
```suggestion
            dependencies.insert(dependency);
```

src/edg.rs
```diff
@@ -372,28 +490,60 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                );
                 self.add_depend(&edge.target, Edges::NEGATION(edge.clone()));
-                let (weight1, weight2) = weight
-                    .split()
-                    .unwrap_or_else(|_| panic!("Worker {} ran out of weight", self.id));
-                self.broker
-                    .send(self.id, Message::NEGATION(edge.clone(), weight1));
-                self.explore(&edge.target, weight2);
+                self.queue_negation(edge.clone(), weight.clone());
+                self.explore(&edge.target, weight.clone());
```
jenrik, line 501: the same weight is used twice. It should be `.split()` to avoid introducing extra into the system

src/edg.rs
```diff
@@ -372,28 +490,60 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
         }
     }
 
+    /// Queueing unsafe negation, which will be queued to negation channel whenever
+    /// release negation is called
+    fn queue_negation(&mut self, edge: NegationEdge<V>, weight: Weight) {
+        let mut len = self.unsafe_edges.len();
+        let mut dist = 0;
+        if let Some(n) = self.distances.get(&edge.source) {
+            dist = n.clone();
```
jenrik, line 526: Fixes `clippy` warning
```suggestion
            dist = *n;
```

src/edg.rs
```diff
@@ -372,28 +490,60 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+        }
+        while len <= dist as usize {
+            self.unsafe_edges.push(Vec::new());
+            len = len + 1
```
jenrik, line 530: Fixes `clippy` warning
```suggestion
            len += 1;
```

src/edg.rs
```diff
@@ -372,28 +490,60 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
 
+    /// Queueing unsafe negation, which will be queued to negation channel whenever
+    /// release negation is called
+    fn queue_negation(&mut self, edge: NegationEdge<V>, weight: Weight) {
+        let mut len = self.unsafe_edges.len();
+        let mut dist = 0;
+        if let Some(n) = self.distances.get(&edge.source) {
+            dist = n.clone();
+        }
+        while len <= dist as usize {
```
jenrik, line 528: this might be better written as a for range loop

src/edg.rs
```diff
@@ -417,7 +567,23 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
                         },
                     );
                 }
-                VertexAssignment::UNDECIDED => {} // fallthrough
+                _ => {
+                    // update distance
+                    let dist = self.distances.get(vertex).unwrap_or(&0);
+                    self.distances
```
jenrik, line 573: Clippy complains about illegal borrow

src/edg.rs
```diff
@@ -417,7 +567,23 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                        .insert(vertex.clone(), max(dist.clone(), distance));
```
jenrik, line 574: Fixes `clippy` warning
```suggestion
                        .insert(vertex.clone(), max(dist, distance));
```

src/edg.rs
```diff
@@ -417,7 +567,23 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+
+                    self.mark_interest(vertex, requester);
+
+                    if let Some(dependencies) = self.depends.get(vertex) {
```
jenrik, line 578: Is it a bug that `dependencies` isn't used or could this have been replaced with
```suggestion
                    if self.depends.contains_key(vertex) {
```

src/edg.rs
```diff
@@ -417,7 +567,23 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
+                        if let Some(assignment) = self.assignment.get(vertex) {
+                            match assignment {
+                                VertexAssignment::UNDECIDED => self.explore(vertex, weight.clone()),
+                                _ => {}
+                            }
```
jenrik, line 583: Fixes `clippy` warning
```suggestion
                            if let VertexAssignment::UNDECIDED = assignment {
                                self.explore(vertex, weight.clone());
                            }
```

src/edg.rs
```diff
@@ -442,43 +609,60 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
...
-                self.broker.send(
-                    *worker_id,
-                    Message::ANSWER {
-                        vertex: vertex.clone(),
-                        assignment,
-                        weight: split_weight,
-                    },
-                )
+        let prev_assignment = self.assignment.insert(vertex.clone(), assignment);
+        let changed_assignment = (prev_assignment != Some(assignment));
```
jenrik, line 613: Fixes `clippy` warning
```suggestion
        let changed_assignment = prev_assignment != Some(assignment);
```

jenrik (changes requested):
Look mostly good

NicEastvillage (approved):

src/edg.rs
```diff
@@ -350,6 +459,21 @@ impl<B: Broker<V> + Debug, G: ExtendedDependencyGraph<V> + Send + Sync + Debug,
 
     // Mark `dependency` as a prerequisite for finding the final assignment of `vertex`
     fn add_depend(&mut self, vertex: &V, dependency: Edges<V>) {
+        // Update the distance
+        let default = 0;
+        let tdist = self.distances.get(vertex).unwrap_or(&default);
+        match dependency.clone() {
+            Edges::NEGATION(edge) => {
+                let sdist = self.distances.get(&edge.source).unwrap_or(&default) + 1;
+                self.distances.insert(vertex.clone(), max(sdist, *tdist));
```
jenrik, line 468: Please fix the `clippy` warning about an illegal borrow

jenrik (changes requested):
there are still `clippy` warnings that needs fixing. Also there is an illegal borrow that absolutely need a fix before we merge

GeneralDenmark (approved):
Looking good :tm:

PR #38, opened 2020-11-23
Intermediate Representation

NicEastvillage:
Adds the next phase of LCGS interpreter. In this phase all symbols are registered and their validity is checked. The result is a symbol table holding all declarations, that are more compact than the original AST versions, as well as lists of each type of declaration (e.g. list of all labels) to allow for faster lookup.

Stuff NOT included in this PR:
- Proper compiler errors. Currently the program will just crashes and gives a sometimes vague description of the problem.
- Relabelling. Requires a relabeler. To get around this you can create a template for each player.
- Implemention of `GameStructure` for the itermediate representation. The implementation will big enough to deserve a PR for itself.


NicEastvillage:
Ignore the `Evaluator`. It is not used atm, but will be later.

GeneralDenmark (approved):
I approve. Det bliver godt, når vi engang kan begynde at kører vores eksempler igennem!

Stor tommelfinger opad.

cogitantium (approved):

PR #53, opened 2020-11-26
Benchmark model checker with cargo bench

jenrik:
Use `cargo bench` to benchmark the model checker om various test cases. Requires the nightly compiler because the `test` feature is not yet stable.

Feel free to contribute test cases by committing to this branch.

### DIFF
--- a/src/edg.rs
+++ b/src/edg.rs
@@ -456,8 +456,15 @@ mod test {
 
     use serde::export::Formatter;
 
+    use crate::atl::dependencygraph::{ATLDependencyGraph, ATLVertex};
+    use crate::atl::formula::Phi;
+    use crate::atl::gamestructure::EagerGameStructure;
     use crate::common::{Edges, HyperEdge, NegationEdge, VertexAssignment};
     use crate::edg::{distributed_certain_zero, ExtendedDependencyGraph, Vertex};
+    use pom::set::Set;
+    use std::hint::black_box;
+    use std::sync::Arc;
+    use test::Bencher;
 
     #[derive(Hash, Clone, Eq, PartialEq, Debug)]
     struct ExampleEDG {}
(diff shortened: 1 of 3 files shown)


src/edg.rs
```diff
@@ -1072,4 +1079,23 @@ mod test {
...
+    #[bench]
+    fn bench_mexican_standoff(b: &mut Bencher) {
+        b.iter(|| {
+            let json = String::from_utf8_lossy(include_bytes!("../mexican-standoff.json"));
+            let game_structure: EagerGameStructure = serde_json::from_str(json.to_str()).unwrap();
+            let formula = Arc::new(Phi::Next {
+                players: vec![0],
+                formula: Arc::new(Phi::Proposition(0)),
+            });
+            let graph = ATLDependencyGraph { game_structure };
```
NicEastvillage, line 1092: We might want to move this out of `b.iter` to avoid measure IO and setup.

jenrik: agreed, I just had some trouble with ownership and didn't bother finding a solution

NicEastvillage: I tried too. Same conclusion xD

NicEastvillage: I tried making the exact same closure in a #[test] and it worked perfectly, so I think it is a problem with how #[bench] modifies the function. I suggest moving to Criterion.rs

cogitantium: Agreed, I'm working on setting up criterion - so far it looks pretty slick:
```
Benchmarking fib 20
Benchmarking fib 20: Warming up for 3.0000 s
Benchmarking fib 20: Collecting 100 samples in estimated 5.0856 s (207k iterations)
Benchmarking fib 20: Analyzing
fib 20                  time:   [25.321 us 25.613 us 25.959 us]
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) high mild
  12 (12.00%) high severe
```

jenrik:
Superseded by #64

PR #52, opened 2020-11-26
Allow LCGS as input game structure

NicEastvillage:
This PR makes it possible to input an LCGS program to the algorithm. I.e. it is now possible to do
```
cargo run -- solver --formula test-formula.json --model test.lcgs --model-type lcgs
```
and have it check if the formula in `test-formula.json` is satisfied in the initial state of the game structure described by `test.lcgs`.

There are still many bad or missing error messages, and the `graph` command only works for eager game structures atm.

Resolves #51 . Depends on #49 .

PR #55, opened 2020-11-30
[LCGS] Adds relabeling feature

### LINKED ISSUE #54: [LCGS] Make relabeling possible
Currently it is not possible to relabel things in templates, which can result in a lot of duplicate code.

NicEastvillage:
Adds the relabeling of templates. In the following block, both John and Bob are workers, but their income is different, as well as when they can work. The words occurances of the given words are simply replaced with their new meaning.
```
player john = worker [income=200, when=1]
player bob = worker [income=300, when=money < 1000]

template worker
    money : [0 .. 10000] init 2000;
    money' = money + income * work - 100;
    [work] when;
    [relax] 1;
endtemplate
```
It is also possible to replace a part of an identifier. In our good old Mexican standoff example we set we `target1` and `target2` to `p2` and `p3`. This works because `p2` and `p3` have the required fields:
```
const max_health = 1;

player p1 = shooter [target1=p2, target2=p3];
player p2 = shooter [target1=p3, target2=p1];
player p3 = shooter [target1=p1, target2=p3];

template shooter

    health : [0 .. max_health] init max_health;
    health' = max(health - (target1.shoot_left + target2.shoot_right), 0);

    label alive = health > 0;

    [wait] true;
    [shoot_right] health > 0 & target1.health > 0;
    [shoot_left] health > 0 & target2.health > 0;

endtemplate
```

Depends on #52 and #49. Resolves #54 .


cogitantium (approved):
Nice :ok_hand:

MathiasMehl (approved):

PR #56, opened 2020-11-30
Replace log4rs with Tracing

jenrik:
replaces log4rs with [Tracing](https://tracing.rs/tracing/)


cogitantium (approved):
Code-wise it looks good but I'm not sure how to use tracing

src/edg.rs
```diff
@@ -554,6 +556,7 @@ mod test {
 
     #[ignore]
     #[test]
+    #[ignore]
```
NicEastvillage, line 559: Very ignored

cogitantium: Much pass

NicEastvillage (approved):

PR #57, opened 2020-11-30
[ATL] Adds more formula term (And, Eventually, Invariant, ...)

NicEastvillage:
This PR adds the following terms to ATL formulas:
* True
* False
* And
* DespiteNext (and renames Next to EnforcedNext)
* EnforceEventually
* DespiteEventually
* EnforceInvariant
* DespiteInvariant

It should be easy to convince yourself that the edges and successor configurations are correct (It's just proving it, that is hard).

There are also some clean up like improving comments and replacing some u32 with usize.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,4 @@ log4rs = "0.12.0"
 serde = { version = "1.0.117", features = ["derive", "rc"] }
 serde_json = "1.0.59"
 lazy_static = "1.4.0"
+joinery = "2.0.0"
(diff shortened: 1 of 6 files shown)


NicEastvillage:
Just realized we are missing True and False. Will do these later.

NicEastvillage:
Done.

Bonus info: The formula `⟪0,1⟫(('1' ∨ ¬('2')) 𝑼 false)` is written like this in json:
```json
{
  "enforce until": {
    "players": [0,1],
    "pre": {
      "or": [
        {
          "proposition": 1
        },
        {
          "not": {
            "proposition": 2
          }
        }
      ]
    },
    "until": "false"
  }
}
```

cogitantium (approved):

NicEastvillage:
Have renamed always to invariant

lars4509 (approved):
Looks good to me

PR #58, opened 2020-12-01
[CLI] Infer model type

NicEastvillage:
This PR makes it possible to omit the `--model-type=MODEL_TYPE` argument if it can be infered from the input models extension. E.g. `standoff.lcgs` will be parsed as LCGS.

PR #63, opened 2020-12-02
Cleans main.rs and fixes the "graph" subcommand

jenrik:


src/main.rs
```diff
@@ -57,120 +58,149 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            let formula = load_formula(formula_path);
 
-    let v0 = game_structure.initial_state_index();
+            handle_json(graph, formula)
+        }
+        "lcgs" => {
+            let lcgs = parse_lcgs(&content).unwrap_or_else(|err| {
+                eprintln!("Couldn't open input model\n\nError:\n{}", err);
+                exit(1);
+            });
```
NicEastvillage, line 188: This error would be a parsing error. File IO errors are handled above.

src/main.rs
```diff
@@ -57,120 +58,149 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            let game_structure = IntermediateLCGS::create(lcgs).unwrap_or_else(|err| {
+                eprintln!("Couldn't open input model");
+                exit(1);
+            });
```
NicEastvillage, line 192: This error would be a semantics error. The user wrote something invalid.

NicEastvillage (changes requested):
Nice! This is much better. Only 8 lines of of dublicate code.

You might now be done with this PR from what I can tell, but here are some thoughts from me:
* This PR overshadows PR #58. Can you including infering of model type in this PR please?
* I would prefer if you wrote "Could not" or even better "Failed to" instead of "Couldn't" in the error messages.
* Comments are missing

NicEastvillage (approved):

GeneralDenmark (approved):
Looking Good :tm:

PR #43, opened 2020-11-24
Examples in our language ☭

GeneralDenmark:
Fixes #25 

Examples which compile in LCGS and at least one working formula:
- [ ] Mexican standoff

PR #64, opened 2020-12-04
Criterion.rs benchmarking

cogitantium:


cogitantium:
Missing proper configuration to access model_check main functions, related to second point in docs: https://bheisler.github.io/criterion.rs/book/user_guide/known_limitations.html

cogitantium:
We should also have benchmarks for important internal functions, such as parsing LCGS

cogitantium:
Now depends on #43

cogitantium:
We've abandoned this approach for benchmarking our tool vs PRISM, instead using https://github.com/d702e20/bench-solver/.

This may become useful in future for micro benching internally for performance improvements.

jenrik:
I think this is still worth keeping open for use in future development

cogitantium:
Codebase adopted by #101, keep this closed until #101 is resolved.

PR #65, opened 2020-12-07
Fixes commandline argument values being read from the wrong struct

jenrik:
The arguments belongs to the subcommand and therefore needs to be read from the subcommand.


cogitantium (approved):
LGTM

NicEastvillage (approved):
Clippy is a bit sad, but I am happy. It works!

PR #69, opened 2020-12-08
[Bug] Fix VarsIterator and DeltaIterator

NicEastvillage:
I printed the graph the mexican standoff and noticed some wrong branches. That lead me to checking the VarsIterator and DeltaIterator. The VarsIterator had a silly mistake, but the DeltaIterator needed recursion to work probably, so I had to do a small rewrite (and introduced PartialMoveIterator in the process). After these changes the Mexican standoff graph looks more as I expect it.


MathiasMehl (approved):

cogitantium (approved):
LGTM

PR #71, opened 2020-12-09
[Bug] Fix use of invert players in despite until

NicEastvillage:
When I was finishing the last case in the proof, I ran into a problem. It turns out of graph is slightly wrong. The despite formulas is NOT supposed to use the inverted players. I checked supervisors Magi 2.0 mail, and he did also not invert the players. That must have been some leftover from a previous attempt. This fixes it and I have triple checked that both the graph in the paper and the algoritm is correct in regards to despite formulas.


jenrik (approved):
Please rename the test files from `happy` to `TRUE`

cogitantium (approved):

GeneralDenmark (approved):
Det er bare shuper

PR #88, opened 2021-01-06
[P8-45/Errors] Add Errors to symbolic analysis phase of LCGS

NicEastvillage:
The LCGS compiler will now produce errors during the symbolic analysis phase. That includes
* Unknown identifiers
* Misuse of identifiers
* Misuse of relabeling

The parsing still doesn't produce errors and the errors does not include line number (yet). This requires rewrite of the parsing.


src/lcgs/ir/symbol_checker.rs
```diff
@@ -47,19 +52,19 @@ impl<'a> SymbolChecker<'a> {
...
+    /// Checks and evaluates an expressions. This is only used in [CheckMode::Const]
     /// where we assume the expression can be reduced to a value already during symbol checking.
-    pub fn check_eval(&self, expr: &Expr) -> Result<i32, ()> {
+    pub fn check_eval(&self, expr: &Expr) -> Result<i32, SymbolError> {
         let checked = self.check(expr)?;
         if let ExprKind::Number(n) = checked.kind {
             Ok(n)
         } else {
-            Err(())
+            panic!("Constant expression was not reduced to a single number.")
```
jenrik, line 62: Shouldn't this be an error?

jenrik (approved):
Instead of printing the Error trait, it might be more idiomatic to implement the Display trait on the Error type and print that instead.

The core functionality of the PR looks good

NicEastvillage:
> it might be more idiomatic to implement the Display trait on the Error type and print that instead.

I did. `src/lcgs/ir/error.rs` line 12. What do you mean?

src/lcgs/ir/symbol_checker.rs
NicEastvillage: Nope. The previous phase should have caught this error already. If it is not a single number here, something is wrong within our compiler.

jenrik:
@NicEastvillage I don't know how I missed that, well then that's all good then

cogitantium (approved):

jenrik (approved):

PR #90, opened 2021-01-07
[P8-49] ATL formula parser

jenrik:
Parses ATL formula into a `Phi` data structure


src/atl/formula/parser.rs
```diff
@@ -0,0 +1,653 @@
...
+    }
+
+    #[test]
+    fn next_1() {
+        assert_eq!(next(&convert).parse(b"X true"), Ok(Phi::True))
+    }
+
+    #[test]
+    fn next_2() {
+        assert_eq!(next(&convert).parse(b"Xtrue"), Ok(Phi::True))
```
NicEastvillage, line 331: I am not sure this should be allowed. Same goes for the other operators. But it's a small thing.

NicEastvillage (approved):
Nice

jenrik:
The `convert` function needs to be split into a function for converting players and a function for propositions.

NicEastvillage:
Yeah, I think we should introduce a trait with two functions: `parse_player` and `parse_proposition`

NicEastvillage:
This feature is done and waiting for reviews from @d702e20/d702e20

src/atl/formula/parser.rs
```diff
@@ -0,0 +1,827 @@
...
+use pom::parser::{list, one_of, seq, sym};
+
+use super::Phi;
+
+/// Parse an ATL formula
+pub(crate) fn parse_phi<'a, 'b: 'a, A: ATLExpressionParser>(
+    expr_parser: &'b A,
+    input: &'a str,
+) -> Result<Phi, String> {
+    let formula = ws() * phi(expr_parser) - end();
```
jenrik, line 14: We should allow whitespace at the end of a formula like we do at the beginning
```suggestion
    let formula = ws() * phi(expr_parser) - ws() - end();
```

src/atl/formula/parser.rs
```diff
@@ -0,0 +1,827 @@
...
+}
+
+/// Allows a CGS model to define custom player and proposition expressions. For instance,
+/// in LCGS we want to be able to write "p2" as a player and "p2.alive" as a proposition, while
+/// in json, players and propositions are numbers.
+pub trait ATLExpressionParser {
+    /// A parser that parses a player name
+    fn player_parser(&self) -> Parser<u8, usize>;
+    /// A parser that parses a proposition name
+    fn proposition_parser(&self) -> Parser<u8, usize>;
```
jenrik, line 27: This should return `Parser<u8, crate::common::Proposition>`. `Proposition` is a alias for `usize` but having a use-case specific name for it helps if we ever need to refactor.

src/lcgs/ir/intermediate.rs
```diff
@@ -463,10 +484,70 @@ impl GameStructure for IntermediateLCGS {
...
+        let parser = identifier() + (sym(b'.') * identifier()).opt();
+        parser.convert(move |(name_or_owner, name)| {
+            // We infer whether the label should be found in local (player) or global scope.
+            // full_name is used for error descriptions.
+            let (symbol, full_name) = if let Some(name) = name {
+                let owner = name_or_owner;
+                (
+                    Owner::Player(owner.clone()).symbol_id(&name),
+                    format!("{}.{}", owner, name),
+                )
```
jenrik, line 522: Why are these lines in parenthesis?

NicEastvillage: Becase the block returns a tuple consisting of a SymbolIdentifier and a String. See line 517 - it's a let statement.

src/atl/formula/parser.rs
NicEastvillage: Good spot. Fixed in 0d67118

src/atl/formula/parser.rs
NicEastvillage: Good idea. Fixed in 0d67118

jenrik: Now that I think about, `player_parser` should probably return `crate::common::Player`

lars4509 (approved):
Looks fine to me, i just have the single comment.

src/main.rs
```diff
@@ -279,6 +262,39 @@ fn load_formula<A: ATLExpressionParser>(
     return (raw_phi, phi);
```
jenrik, line 262: Unneeded `return` statement

jenrik (commented):
Highlighted some Clippy warnings

docs/atl-grammar.txt
```diff
@@ -0,0 +1,45 @@
...
+
+despite_next := despite_players next
+despite_until := despite_players until
+despite_eventually := despite_players eventually
+despite_invariant := despite_players invariant
+
+proposition := "'" identifier "'"
+not := "!" phi
+or := phi "|" phi
+and := phi "&" phi
```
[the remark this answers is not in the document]
MathiasMehl: I see your point, but I am in a different mindset when writing ATL formulae as opposed to LCGS, so I personally don't confuse the syntax. 
I'm fine with either solution (changing or not).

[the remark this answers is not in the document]
jenrik: I would rather that we change LCGS, since the formula syntax is based on a previously established syntax for LTS formulae that is also used elsewhere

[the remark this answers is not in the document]
MathiasMehl: I assume you mean LTL? AFAIK LTL uses  ∧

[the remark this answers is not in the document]
NicEastvillage: Depends on if we ever want to support bitwise operators in LCGS.

lars4509 (approved):

MathiasMehl (approved):

PR #91, opened 2021-01-08
[P8-47] Slightly better parse errors when parsing LCGS

NicEastvillage:
The parse errors now include which on line the erroneous declaration begins. However, if the error is inside a template, the error will still point to the beginning of the template. Proper errors requires a big rewrite of the parser, which we don't have time for before the exam.


lars4509 (approved):

test.json
```diff
@@ -0,0 +1,3 @@
+{
+  "proposition": 0
+}
```
cogitantium, line 3: I don't think you meant to include these test.(json/lcgs) files

cogitantium (changes requested):

test.json
NicEastvillage: Right. Fixed in 827281b

cogitantium (approved):

PR #92, opened 2021-01-12
[P8-48] Fix successors of invariant formulas

### LINKED ISSUE #83: Wrong result when testing p1_p2_remain_stupid_together_TRUE.json
The formula should return TRUE, but returns False.

The formula tests whether or not p1 and p2 can cooperate to them both never being omniscient.
They should be able to do this, because they can just keep calling each other, and never getting secret 3 and 4, which is needed to be omniscient

NicEastvillage:
In the previous implementation cycles would always become false, even if all states in the cycle satisfied the subformula. So I swapped it with the definitions of the paper:
billede

I think this fixes #83

Tested using the graph command.

### DIFF
--- a/src/atl/dependencygraph.rs
+++ b/src/atl/dependencygraph.rs
@@ -161,12 +161,10 @@ impl<'a> Iterator for PartialMoveIterator<'a> {
         if !self.initialized {
             self.make_first();
             Some(self.current.clone())
+        } else if self.make_next(0) {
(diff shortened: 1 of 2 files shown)


jenrik (approved):

GeneralDenmark (approved):
Små kode forbedringer :+1: 
LGTM

cogitantium (approved):

PR #93, opened 2021-01-15
Remove [ignored] from working tests

NicEastvillage:
All of our tests now succeeds, so let's remove the [ignore] annotation so we can keep it that way.


jenrik (changes requested):
Does not compile 👎 
There is some broken use statements referring to serde that should probably refer to the std library instead.

NicEastvillage:
I am aware. Somehow master compiles fine. But I fixed the serde import on the more important #92 PR

cogitantium (approved):

jenrik (approved):

PR #95, opened 2021-02-19
[fix-readme-lcgs-example] fixes lcgs example in the readme

MathiasMehl:
The example used the wrong syntax for logical and (& instead of &&) and used "true" which we do not have in the language.


NicEastvillage (approved):

README.md
```diff
@@ -29,9 +29,9 @@ template cowboy
 
     label alive = health > 0;
 
-    [wait] true;
-    [shoot_right] health > 0 & target1.health > 0;
-    [shoot_left] health > 0 & target2.health > 0;
+    [wait] 1;
```
lars4509, line 32: You could probably still have true here, and make a const true = 1, in order to still have the same readibility.

lars4509 (approved):
Looks good, have one small suggestion of little importance.

README.md
MathiasMehl: That's true (haha), but I think most programmers will understand 1 as meaning true in this context, so for a short example I don't think it is necessary

PR #100, opened 2021-02-25
[P8-63] Add Tic Tac Toe lcgs

NicEastvillage:
Tic Tac Toe, a turn-based game between nought and cross. Each turn a the given player adds a piece of their shape to an empty spot on the 3x3 grid. If a player manages to get 3 of their shapes on a straight line, they win. The player with crosses starts. It has been proven, that if there exists a strategy for the starting player such that they cannot lose.

(This implementation is the variant, where pieces are not removed again)

ATL quries include:
* Can cross avoid losing (TRUE expected): `<<cross>> F !nought_wins`
* Can cross guarantee winning (FALSE expected): `<<cross>> F cross_wins`
* Can nought avoid losing (TRUE expected): `<<nought>> F !cross_wins`
* Can nought guarantee winning (FALSE expected): `<<nought>> F nought_wins`


docs/lcgs/working-examples/Tic_Tac_Toe/tic_tac_toe.lcgs
```diff
@@ -0,0 +1,118 @@
...
+    [play13] !board_full && !winner && turn == SHAPE && pos13 == EMPTY;
+    [play21] !board_full && !winner && turn == SHAPE && pos21 == EMPTY;
+    [play22] !board_full && !winner && turn == SHAPE && pos22 == EMPTY;
+    [play23] !board_full && !winner && turn == SHAPE && pos23 == EMPTY;
+    [play31] !board_full && !winner && turn == SHAPE && pos31 == EMPTY;
+    [play32] !board_full && !winner && turn == SHAPE && pos32 == EMPTY;
+    [play33] !board_full && !winner && turn == SHAPE && pos33 == EMPTY;
+endtemplate
+
+label tie = board_full && winner == EMPTY;
```
lars4509, line 116: In this case we have a tie, when board not full and we have a winner. Which seems strange to me.

lars4509 (approved):
Looks fine to me. Have a single comment on definition of tie.

docs/lcgs/working-examples/Tic_Tac_Toe/tic_tac_toe.lcgs
NicEastvillage: I don't understand. The game is tied if
* the board is full (no empty positions), and
* there's no winner (no one got three of their shapes on a row)

lars4509: I go confused by precedens. I read it like `(board_full && winner) == EMPTY`. Now i see that is not the case. Then no problems :)

MathiasMehl:
Looks good to me. Could add a formula testing if a player can guarantee a tie

MathiasMehl (approved):

PR #101, opened 2021-03-01
[P8-53] Profiling research, documentation, and implementation

cogitantium:


cogitantium:
When/if Criterion benchmarking suite becomes our main metric, we can use an Github Action to display at-a-glance benchmark stats against any branch (main or specific target). Results would fluctuate with load on Github but if not useful, they're at least cool.
image
Regardless, when measuring time (Criterion) we need to use dedicated, quiet machines, so we would always need at least our own machines or optimally a dedicated job on the MMC.

jenrik:
Why are we implementing our own profiling functionality through some counting of magic strings? This seems like a cumbersome approach and also have issues with congestion on access to stdout. A normal sampling profiler should work is simpler to use, and gives good enough results. Below is an image of what the sampling profiler in Microsoft Visual Studio can give when run on our program:
unknown

I am pretty sure there exists equivalent sampling profilers for Linux and Mac. They should also be able to give functions names since that can be annotated via symbols.

cogitantium:
@jenrik yes, IDE tools also exist. Note that this PR is still draft. CLion seems to have the same capabilities as you show:
image

The counts-functionality is not meant as a replacement but may be useful if people find built-in profiling tools difficult to work with or have very specific needs for their testing. Furthermore, we can include internal values in this printing which may prove useful. You should read the 'Worse is better' chapter of the Counts readme: https://github.com/nnethercote/counts/.

cogitantium:
Tests fail because access had to be changed for Criterion to access dependency graph, etc. I've validated that tests work and benchmarks fail, when these attempts are reverted, the inverse is true at 10e17a9. I might need some help to solve this elegantly.

NicEastvillage:
It looks like the tests fail because you haven't updated the imports in one or more of the test modules.

NicEastvillage (approved):
Pretty cool. Only missing measurement is time spent in each phase of the worker and maybe event counts per worker. But this definitely suffices for now! I did not find the cause of the test errors though. I will have to look deeper.

MathiasMehl (approved):

PR #89, opened 2021-01-06
[P8-46/Errors] Cleaner handling of errors in main

NicEastvillage:
Cleaner handling of errors in main. The most noteworthy change is the introduction of the `ModelType` enum to avoid checking string equality.

Depends on #88

PR #102, opened 2021-03-01
Add exit 1 instead of panic on unwrap of null args

cogitantium:
Does not return the help-message on no-args but exits cleanly with code 1. I think this is sufficient as there's no obvious way to make clap emit its help-message.


NicEastvillage (approved):
Approved.
Note, I made some additional clean ups in #89, which still hasn't been reviewed.

jenrik (changes requested):
https://docs.rs/clap/2.33.3/clap/struct.App.html#method.print_help

NicEastvillage:
Note that the modified line has been deleted in PR #90. You might want to check that out and/or rebase from there.

cogitantium:
@jenrik using https://docs.rs/clap/2.33.3/clap/struct.App.html#method.print_help is not a viable route as App is consumed upon using get_matches, making it difficult to call print_help after using the result of get_matches to check if subcommands exist.

@NicEastvillage and I tried to solve it by using the setting AppSettings::ArgRequiredElseHelp (https://docs.rs/clap/2.31.1/clap/enum.AppSettings.html#variant.ArgRequiredElseHelp), however there's issues with subcommands counting as arguments and if any default values are set (e.g. log-level), this effectively negates this setting.

The changes initially suggested is an easy fix and since this issue generates very little value for the time it takes, I think we should save it for later. I'll make an issue.

src/main.rs
```diff
@@ -77,6 +77,32 @@ fn main_inner() -> Result<(), String> {
     setup_tracing(&args);
     trace!(?args, "commandline arguments");
 
+    let subargs = args.subcommand().1.unwrap_or(exit(1));
```
jenrik, line 80: Check: Unreachable call

jenrik (changes requested):
Something is fishy, `cargo check` complains that line 80 is unreachable

src/main.rs
NicEastvillage: Ah yes. `exit(1)` is NOT lazily evaluated, so the program stops here.

PR #103, opened 2021-03-02
[BUG] Fix premature i32 conversion in LCGS state translation

NicEastvillage:
@MathiasMehl stumbled upon a bug. I found that it originated in our LCGS state translation functions. Turns out the carry is an i32 and therefore loses precision when coming from usize. Instead, the conversion to i32 should happen when converting the remainder to a variable value.


MathiasMehl:
Did some testing on the Peterson lcgs programs, they now behave sanely and as intended. (did not test on 09 and 10 because they take a long time to test)

MathiasMehl (approved):

lars4509 (approved):
I think i understand the problem and the fix. Looks good to me.

src/lcgs/ir/intermediate.rs
```diff
@@ -76,11 +77,11 @@ impl IntermediateLCGS {
             let symb = self.symbols.get(symb_id).unwrap();
             if let DeclKind::StateVar(var) = &symb.kind {
                 let value = {
-                    let size = var.ir_range.end() - var.ir_range.start() + 1;
+                    let size = (var.ir_range.end() - var.ir_range.start() + 1) as usize;
```
lars4509, line 80: I might ask stupid but i am not sure why this is `as usize` and not `as i32` maybe one can elaborate?

NicEastvillage: `end()` will always be greater than `start()`, so we are guaranteed to get a positive number, but I then also cast it to `usize` since I a few lines later do an operation, where both operands must be of the same type. The other operand is of type `usize` and since `usize` > `i32` we cast upwards to `usize` to not lose precision.

Issue #104, opened 2021-03-02
features for atl parser


MathiasMehl:
Allow expressions in atl formulae, such as `<<1,2>> F (1.health < 2)`

Be able to say "all players" in the path quantifiers, such that if i have 5 players, i don't have to say e.g `<<1,2,3,4,5>>` but instead be able to say `<<\Sigma>>`

NicEastvillage:
I think `Exists` and `Always` is better - or at least something we should have too. They simple replace the player coallition, e.g. `Exists F proposition`

PR #99, opened 2021-02-25
More comments on working examples

MathiasMehl:
I added some comments to the cases that I have made, that needed more comments to be understood. 
Some other of our cases still need a bit of documentation to be easily understood.

PR #107, opened 2021-03-04
[P8-65] Add peterson lcgs

MathiasMehl:
Adds lcgs implementation of the peterson algorithm. Note that the implemented version is the Filter algorithm, which can be used when there are more than 2 processes (original peterson is limited to 2 processes), see https://en.wikipedia.org/wiki/Peterson%27s_algorithm.

Versions with 2-9 processes are added, in their own directory. In their directory, there can also be found 3 formulae, in both a json and atl version. If you want to look through the implementation of the algorithm, just look in the peterson-02.lcgs, the versions with more processes are similar, just scaled to more processes.

`ensure-mutual-exclusion-TRUE-02` tests if no matter what the players choose to do, there will always be at most 1 process in the critical section. ATL version is `[[p0,p1]] G (mutual_exclusion)`. The json version uses `despite invariant`, which at the moment is buggy and returns wrong result, so perhaps don't look to closely on this one. 

`multiple-in-CS-FALSE-02` tests if we at any point can reach a state with more than one process in the critical section. Very similar to above, but uses `enforce eventually` instead in the json version, returns correct result. ATL version `<<p0,p1>> F (multiple_in_cs)`. 

`p0-eventually-reach-CS-TRUE-02` tests if the process p0 can eventually reach the critical section. ATL version `<<p0,p1>> F (p0_in_cs)`. 

The labels used in the ATL versions can be seen first in the peterson-0*.lcgs

Right now it introduces ALOT of files, this can be reduced by fixing issue #104 so there wont be many very alike files.


NicEastvillage:
PR #99 has been merged into main. Please merge main into your branch and move the Peterson LCGS to the `lcgs-examples` folder.

lars4509 (approved):
Looks good to me.
While this problem is interesting, it might not be as interesting in the ATL perspective, since all formulas are affecting all players and in that case could be expressed with CTL (correct me if i am wrong). That said it still might suit as a good test case.

NicEastvillage (approved):

PR #110, opened 2021-03-05
[P8-74] Minimum solve set algorithm

NicEastvillage:
The `minimum_solve_set` algorithm will find the smallest set of vertices that need to be checked to find the assigment of every vertex in an extended dependency graph assuming a local algorithm with a perfect heuristic. The algorithm uses a recursive approach since a perfect heuristic would induce a depth-first search that only explores the vertices needed to confirm the root's assignment.

Note that found solve sets are minimum, but not necessarily the only minimum solve set.

With this algorithm we can analyse which branches of specific formulas usually have the least/highest solve distance, which brings us one step closer to constructing a heuristic that can prioritise some edge over others.

This PR depends on #114


NicEastvillage:
Note to future me. There is an ambiguity in the current result as it depends on the edge order in `succ`. For instance, the vertex C in this EDG has two valid answers.
```rust
#[test]
fn test_sd_complicated_01() {
    simple_edg![
        A => -> {B} -> {C, E};
        B => -> {D};
        C => -> {B};
        D => -> {C} -> {};
        E => .> F;
        F => ;
    ];
    assert_signed_solve_dists!(
        root=A,
        A => 3,
        B => 2,
        C => 3, // FIXME Both 3 and -1 are valid answers
        D => 1,
        E => 2,
        F => -1,
    );
}
```
This ambiguity should be eliminated as there should only be a single answer with a perfect heuristic. However, since this ambiguity  is created by two different enters of a cycle, it is not so straight forward. Damn cycles.

NicEastvillage:
The previously mentioned error can be solved by marked negative assignments caused by cycles, and then always recalculate negative assignments caused by cycles, updating them to either a more negative number or to a positive number.

But. I found another problem with positive hyper-edges, where some vertices are counted multiple times. This problem seems harder to fix:
```rust
#[test]
fn test_sd_complicated_01() {
    simple_edg![
        A => -> {B,C,D};
        B => -> {};
        C => -> {B};
        D => -> {C};
    ];
    assert_signed_solve_dists!(
        root=A,
        A => 4, // FIXME Assigned 7 due to summation, but should be 4
        B => 1,
        C => 2,
        D => 3,
    );
}
```

NicEastvillage:
Both errors has been fixed. Double-counting is solved by switching from solve distance to minimum solve set (the minimum set of vertices that must be checked) and using unions.

Cycle errors are solved by marking false assignments with an uncertainty, which is true, when the assignment is based on hitting a cycle. Uncertain assignments is always recalculated, ensuring that alternative ways to enter the cycle is also correct. There is one caveat though. Stray uncertain vertices are not recalculated, which means they are not correct. However, their relative size to their siblings is correct.

Calculating the correct assignment of uncertain vertices could be done by repeating the algorithm starting from those, but it might be very slow since we have to iterate through all vertices again. I decided not to add that for now.

lars4509 (approved):
Seems good to me

atl-checker/src/main.rs
```diff
@@ -160,6 +159,42 @@ fn main_inner() -> Result<(), String> {
...
+            let input_model_path = analyse_args.value_of("input_model").unwrap();
+            let model_type = get_model_type_from_args(&analyse_args)?;
+            let formula_path = analyse_args.value_of("formula").unwrap();
+            let formula_format = get_formula_format_from_args(&analyse_args)?;
+
+            fn analyse_model<G>(graph: ATLDependencyGraph<G>, v0: ATLVertex)
+            where
+                G: GameStructure + Send + Sync + Clone + Debug + 'static,
+            {
+                println!("Analysing EDG ...");
```
cogitantium, line 172:
```suggestion
                println!("Analysing EDG (state formula: length)");
```

Interpreting output is a bit difficult, when you don't know what you're looking at. I think we should include a small hint like this. Maybe length should be depth instead?

atl-checker/src/main.rs
```diff
@@ -160,6 +159,42 @@ fn main_inner() -> Result<(), String> {
...
+                let mss = minimum_solve_set(&graph, v0);
+                for (vertex, assignment) in mss {
+                    println!("{}: {}", vertex, assignment.len());
+                }
+                println!("Done");
```
cogitantium, line 177:
```suggestion
```

Looks like testprinting, if program exits without panic and with retcode 0, it's done, so no need to dirty stdout.

cogitantium (changes requested):
Well explained algorithm, looks good. Just comments on printing

NicEastvillage:
@cogitantium don't think too hard about the output. It will be refactored soon. I might even remove the analyse subcommand again and call the function from another script instead. In any case the output should be a csv file with this and more data about the edgs.

cogitantium:
@NicEastvillage okay that's cool, we might as well commit the changes then - as it will be *more* correct in the meantime

cogitantium (approved):

PR #112, opened 2021-03-08
[P8-61] Network Topology

lars4509:
In a network topology you can multiple routers which are interconnected, but since hardware can be unreliable we have to ensure that even if multiple routers fail, there is an route from one router to another. These LCGS models depict such topologies with multiple router which can either fail or make new connections. 

This is still only a draft since the current implemented LCGS models for the network topologies are to small to check anything interesting.


lars4509:
I am gonna close this request, since i do not believe it will be actual at any point.

PR #114, opened 2021-03-09
Macro to create and test hardcoded EDGs

NicEastvillage:
This PR introduces macros for defining and testing EDGs. The following macro defines an EDG:
```rust
simple_edg![
    A => -> {B, C} -> {D};
    B => ;
    C => .> D;
    D => -> {};
];
```
Combined with the following macro, we can test the assignment of each vertex:
```rust
edg_assert!(A, TRUE);
edg_assert!(B, FALSE);
edg_assert!(C, FALSE);
edg_assert!(D, TRUE);
```


NicEastvillage:
You know it is a good macro, when it is actually enjoyable converting the old tests. I even added some more tests, because it was so easy!

src/simple_edg.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// edg_assert!(A, TRUE);
+/// edg_assert!(B, FALSE);
+/// ```
+/// We list the edges of each vertex, separating each vertex using a semicolon.
+/// The syntax `-> {V, ..., W}` defines a hyper-edge with targets `V`,...,`W`, and the syntax
+/// `.> P` defines a negation edge with target `P`. Declare multiple edges by writing them
+/// after each other (as seen for vertex `A` and `C` above), but note that hyper-edges must
+/// precede negation edges. Additionally, all targets must also be declared in the first line
+/// of the macro, and the vertex declarations and the edge declarations are separated with `::`.
```
jenrik, line 23: The note about the `::` syntax needs to be removed since we no longer have explicit vertex enumeration

src/simple_edg.rs
```diff
@@ -0,0 +1,158 @@
...
+///     A => -> {B} .> {D};
+///     B => -> {D, C};
+///     C => ;
+///     D => -> {C};
+/// ];
+///
+/// edg_assert!([MyEDG1, MyVertex1] A, TRUE);
+/// edg_assert!([MyEDG1, MyVertex1] B, FALSE);
+/// ```
+/// The `edg_assert` macro allows the same naming functionality.
```
jenrik, line 43: I do not think we need to mention that here

src/simple_edg.rs
```diff
@@ -0,0 +1,158 @@
...
+/// ```
+/// simple_edg![
+///     A => .> B;
+///     B => -> {};
+/// ];
+///
+/// edg_assert!(A, FALSE);
+/// edg_assert!(B, TRUE);
+/// ```
+/// Note that TRUE/FALSE must be capitalized.
```
jenrik, line 107: Good to have this noted down 👍

src/simple_edg.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// # Worker count
+/// You can set the worker count by supplying a third argument to the marco. The default
+/// number of workers are 3.
```
jenrik, line 111: probably wise to not have it default to 1 👍

jenrik (changes requested):
It's very confusing that all the tests have been given generic names.

src/simple_edg.rs
NicEastvillage: I am not sure about that. Testing the multi-threadedness is also important for the `distributed_certain_zero` algorithm. For instance, we have three tests that does not work when worker count = 3, but do when worker count = 1. These tested have been marked with `#[ignore]` in this PR and should hopefully be fixed in #111. But I only found them because I changed our default worker count to 3 (from 1).

src/simple_edg.rs
NicEastvillage: Does not hurt to have.

src/simple_edg.rs
jenrik: Yes, and the discovery of these failing tests is exactly why I think it's good that we have changed the default to 3

NicEastvillage: Oh, I think I misread your comment then :D

jenrik (approved):

lars4509 (approved):
Really cool.

atl-checker/src/simple_edg.rs
```diff
@@ -0,0 +1,158 @@
...
+///     D => -> {};
+/// ];
+///
+/// edg_assert!(A, TRUE);
+/// edg_assert!(B, FALSE);
+/// ```
+/// We list the edges of each vertex, separating each vertex using a semicolon.
+/// The syntax `-> {V, ..., W}` defines a hyper-edge with targets `V`,...,`W`, and the syntax
+/// `.> P` defines a negation edge with target `P`. Declare multiple edges by writing them
+/// after each other (as seen for vertex `A` and `C` above), but note that hyper-edges must
```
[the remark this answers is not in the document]
NicEastvillage: `D` does not have multiple edges though

MathiasMehl:
_Nice_

I however think that we could/should have better naming practices for the tests, as they say little to nothing about what is being tested. (Not to say that it was much better before this PR)

NicEastvillage:
I discussed that with Jener as well. But none of these tests test anything specific. Those that force release of negation edges have been named `test_negations_X`. If you have some good names for the others, I would not mind changing them.

MathiasMehl (approved):

PR #115, opened 2021-03-15
Restructure into a Cargo workspace

jenrik:
This PR moves the Cargo project into a subdirectory, and creates a workspace definition. The changes allows us to add separate components in the future, fx moving the CLI into a crate of its own.

https://doc.rust-lang.org/cargo/reference/workspaces.html


NicEastvillage (approved):

jenrik:
When a new crate is added to the workspace it can use the atl checker by adding the following its to `Cargo.toml`

```toml
[dependencies]
atl-checker = { path = "../atl-checker", version = "0.1.0"}
```

Be warned that the version number must match with the one written in `atl-checker/Cargo.toml`

lars4509 (approved):

PR #116, opened 2021-03-16
More formulae in peterson case

MathiasMehl:
Adds a formula for testing if we can reach a deadlock, and adds another version of the p0 can reach cs formula. 
Same formula as in the fischer case.

The lcgs code in peterson-08 and 09 has kinda exploded in size, because of the label testing for deadlock. In the smaller cases I could just add a variable keeping track of whether or not the process is in a deadlock, but this increases the state space. In the largest cases, I was simply too close to the max number of states, so I couldn't add more variables, and instead make a very large expression in the label instead.


NicEastvillage (changes requested):
You use the proposition `deadlock` in the atl formula, but in the lcgs problem, it is called `process_in_deadlock`.

NicEastvillage (approved):

cogitantium (approved):

PR #125, opened 2021-03-24
[P8-84] Added index translation

NicEastvillage:
Extends the trait GameStructure with translation methods that translates the indexes of stuff to human-readable names. For EagerGameStructure (json), the names are still just numbers, but for the LCGS game structure, the names are the names used in the LCGS source code.

The translation is currently not used.

This resolves P8-84, a child issue to P8-69.


cogitantium (changes requested):
You need to untrack `Cargo.lock`. Otherwise nice

lars4509 (approved):
Looks good

jenrik (approved):

PR #126, opened 2021-03-25
Move receive msg to Broker trait

jenrik:
The current receive logic is hardwired to use `ChannelBroker` receive channels, this PR changes that to into a more generic interface for receiving messages. The will be needed to in order implement different brokers, for instance network based brokers.


NicEastvillage (approved):
I am unsure about moving `terminate` into `return_result` since the main thread sometimes wants to collect other results from the workers (e.g. assignments, so a partial strategy can be calculated). But I guess we tackle that when we need it.

lars4509 (approved):
Very cool!

PR #128, opened 2021-03-27
Make hash function changeable, and use faster hash function

jenrik:
This change makes `ExtendedDependencyGraph` and `Worker` to be generic over the `Hasher` they uses, and swaps the default hash function for [XX hash](https://docs.rs/fasthash/0.4.0/fasthash/xx/index.html) on supported operating systems (anything but Windows).

There is some known bugs:
* hashing of vertices is inconsistent, and therefore we use the default hasher for hashing vertices until a solution have been found.
* running `cargo bench`  yield many `panicked at 'Send to worker 4 failed: "SendError(..)"'` errors.


jenrik:
**Don't trust any benchmarks of this change before the known bugs have been fixed**

falkecarlsen:
Closed because of staleness, reopen if interest arrives.

PR #130, opened 2021-03-29
[P8-86] Support for search strategies

NicEastvillage:
This PR adds supports for search strategies for the distributed certain zero. The search strategy manage in which order safe hyper-edges and negation edges are processed. Currently, there is only one search strategy: breadth-first search.

Depends on #129 for no reason


cogitantium (commented):
Buena 👌

cogitantium (approved):

MathiasMehl (approved):

PR #131, opened 2021-03-30
[P8-83] collecting of assignments

lars4509:
Collecting all assignments in the main thread for further use


jenrik (commented):
I'm not entirely sure what this i good for. This will definitely become problematic when all workers no longer live in the same process (fx networked workers). Also the overhead of copying all the assignment to the main thread can become very expensive, try see if it instead can be moved.

atl-checker/src/edg.rs
```diff
@@ -53,6 +53,16 @@ pub fn distributed_certain_zero<
...
         .receive_result()
         .expect("Error receiving final assigment on termination");
+
+    let mut assignment_table = HashMap::<V, VertexAssignment>::new();
+    for _ in 0..worker_count {
+        let assignments = manager_broker
+            .receive_assignment()
+            .expect("Error receiving the assignment table");
+
+        assignment_table.extend(assignments);
```
NicEastvillage, line 63: Extending the assignment table overrides values, but certain assignments needs to have priority, i.e. if one worker found a certain assignment, then that must be the one put into the table. Unknown < undecided < {true, false}.

NicEastvillage (changes requested):

NicEastvillage:
@jenrik we want to compute the partial strategy (a partial function from states to moves for each relevant player) that proves that the formula is true/false. We are still not entirely sure how to do this, but having the assignment is the first step.

jenrik:
I still uphold my concerns. I think you should delay this change until you know how to compute a partial strategy

NicEastvillage:
Closing since it is so far behind main. I will resolve the conflicts and open PR, merging it to a development branch.

PR #132, opened 2021-04-01
[P8-87] Make succ return Vec for more structure

NicEastvillage:
Our original plan was to make `succ` return an enum of structs such that we had full insight into the various edges produced by specific formulae as this would aid us create better analysis for heuristics and easier computation of partial strategies.

However, that would have affected the generallity of the algorithm as it forces all vertices to have ATL+CGS-like branches. For instance, our test EDGs are not generated from ATL+CGS, so they would not fit a similar struct. Instead, I chose to replace the HashSet with a Vec and introduce a defined order in `succ`. The order follows how we have previously drawn the successor edges, left-to-right. This order is similar to having the smallest edge or the smallest target first. Due to the nature of the successors, this order only matters for Until and eventually formulae.

As an alternative way to aid computation of partial strategies, this PR adds partial moves to hyper-edges where relevant.

Depends on #130

### DIFF
--- a/atl-checker/src/atl/dependencygraph.rs
+++ b/atl-checker/src/atl/dependencygraph.rs
@@ -311,11 +311,11 @@ impl<G: GameStructure> ATLDependencyGraph<G> {
 }
 
 impl<G: GameStructure> ExtendedDependencyGraph<ATLVertex> for ATLDependencyGraph<G> {
+    /// Produce the edges of the given vertex
+    /// Where possible, the smallest edge will be the first in the produced vector,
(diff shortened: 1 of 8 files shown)


cogitantium (approved):
Nice, looks good - a thought I had regarding the builder pattern; does this allow us use different strategies for workers within a single run? I.e. Half does BFS other half does DFS - I'm not sure if it's beneficial but sounds cool.

NicEastvillage:
Sounds like you are talking about #130. But the answer is yes.

cogitantium:
Yes, I missed it depended on #130 - sound good

MathiasMehl (approved):

PR #140, opened 2021-04-16
[P8-88] Restructuring

NicEastvillage:
This PR restructures everything in the `atl-checker::src` folder and might break many other branches.

The picture below shows the new structure:

billede

Most notable changes:
- Certain zero algorithm (and minimum solve set algo) and related stuff has been moved to `algorithms`. This means search strategies are located at `algorithms::certain_zero::search_strategies`
- Definitions of EDGs remains in `edg`
- ATL stuff is still in `atl` while the CGS (game structure) stuff is moved to a separate `game_structure`. This includes LCGS now in `game_structure::lcgs`

I found one ugly dependency during this restructuring. The `edg::HyperEdge` has a `pmoves` field of type `PartialMove`, creating an assumption that all EDGs are created from ATL formulae. This is something we can get rid of with generics, but it will not be in this PR. Hence, both generic EDG and EDG-created-from-ATL stuff can be found in `edg`.


cogitantium (approved):
Benchmarking still works, nice :+1:

GeneralDenmark (approved):
Bench marking still works :+1:

PR #142, opened 2021-04-19
Allow SearchStrategy to modify edges during exploration

NicEastvillage:
This change allows SearchStrategies to modify the edges. This will allow strategies to optimize the order of targets (linear heuristic will likely do this).

Due to the caching of successors, the edges can only be modified once (during exploration, right after generation), and I had to inline the previous `succ` function to achieve this.

Depends on #140

### DIFF
--- a/atl-checker/src/algorithms/certain_zero/search_strategy/mod.rs
+++ b/atl-checker/src/algorithms/certain_zero/search_strategy/mod.rs
@@ -24,6 +24,12 @@ pub trait SearchStrategy<V: Vertex> {
     fn queue_back_propagation(&mut self, edge: Edge<V>) {
         self.queue_new_edges(vec![edge])
     }
+
+    /// Modify the edges to benefit the search strategy. For instance, this could be a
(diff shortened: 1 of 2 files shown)


atl-checker/src/algorithms/certain_zero/search_strategy/mod.rs
```diff
@@ -24,6 +24,12 @@ pub trait SearchStrategy<V: Vertex> {
     fn queue_back_propagation(&mut self, edge: Edge<V>) {
         self.queue_new_edges(vec![edge])
     }
+
+    /// Modify the edges to benefit the search strategy. For instance, this could be a
+    /// rearrangement of the targets of the edges which results in better performance.
+    fn modify(&mut self, edge: Vec<Edge<V>>) -> Vec<Edge<V>> {
+        return edge;
```
lars4509, line 31:
```suggestion
        edge
```
This should fix the warning

lars4509 (approved):

MathiasMehl (approved):

PR #144, opened 2021-04-23
[P8-83] Collecting of assignments

NicEastvillage:
First step in calculating partial strategy output.

This PR is the new #131, targeting the development branch `partial-strategies`.


NicEastvillage:
I have an idea to make this more inter-process communication friendly: Only send certain assignments, one at a time, in the `final_assign` method, and only if the given worker is the owner of the given vertex. This breaks down the assignments to many small messages AND less data is sent in total, since workers only send their own certain assignments (no uncertain, and none received from other workers).

MathiasMehl (approved):

PR #147, opened 2021-04-28
[P8-113] Add AnnotatedEdges for computation of partial strategies

NicEastvillage:
This PR introduces a new EDG with annotations on edges and their targets. This AnnotatedEDG will be used to compute partial strategies.

The `AnnotatedEDG::succ` is similar to `EDG::succ` except it adds annotations as well. It is separate since Certain Zero does not use the annotations, and they would hence unnecessarily increase the size of messages between workers.

When computing partial strategies, these annotations will be the partial move that resulted in that edge (if relevant). This will make it easy to say "The targets of this edge is true, then it must have been this move".


MathiasMehl (approved):

falkecarlsen:
@NicEastvillage you might want to merge from main at this point, especially to include updated benchmarking and CI flow Github action

NicEastvillage:
> shouldn't this PR point to compute-partial-strategy which is source branch of #148 instead of current target partial-strategies?

No, this is part of the development of partial strategies which happens on the `partial_strategies` branch. `compute-partial-strategy` will also be merged into `partial_strategies` afterwards

PR #149, opened 2021-04-30
Linear optimize

MathiasMehl:
Uses ideas from linear programming to queue next edges when exploring the graph. 
If the Edge is a HyperEdge, iterate through all targets and compute average distance in all target, return this.
If Edge is a NegationEdge, just return distance.
From each target, extract (if possible) the linear expression from the formula, convert this to a linear programming problem, find the current value of the relevant StateVar used in the expression, and find distance from this value to the solution and return this as distance. 

Implements a simple cache using the State and Formula as key to a value
- Could extend this cache

Uses a priority queue with distance as priority

Next step is (fixing) and extending the is_linear and extract linear expression functions, to allow more expressions to be used in the linear programming


atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,515 @@
...
+    }
+}
+
+impl SearchStrategy<ATLVertex> for LinearOptimizeSearch {
+    /// Simply returns the edge with highest priority (i.e lowest distance)
+    fn next(&mut self) -> Option<Edge<ATLVertex>> {
+        let edge = self.queue.pop();
+        if edge.is_some() {
+            Some(edge.unwrap().0)
+        } else { None }
```
NicEastvillage, line 69: This whole block can be replaced with
```rust
self.queue.pop().map(|entry| entry.0)
```

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,515 @@
...
+    fn queue_new_edges(&mut self, edges: Vec<Edge<ATLVertex>>) {
+        for edge in edges {
+            let distance = self.get_distance_in_edge(&edge);
+
+            // Add edge and distance to queue
+            if let Some(dist) = distance {
+                self.queue.push(edge, -dist as i32);
+            } else {
+                // Todo what should default value be, if cannot be calculated?
+                self.queue.push(edge, 0);
```
NicEastvillage, line 83: Yeah, this is a big question/problem because it affects the performance by a lot, but there isn't a correct answer. This is one of the reasons I think we should sort the targets instead of prioritising (incomparable) edges.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,515 @@
...
+                            GreaterThan => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: GreaterThan }) }
+                            LessThan => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: LessThan }) }
+                            GreaterOrEqual => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: GreaterOrEqual }) }
+                            LessOrEqual => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: LessOrEqual }) }
+                            And => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: And }) }
+                            Or => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: Or }) }
+                            Xor => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: Xor }) }
+                            Implication => { Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: Implication }) }
+                        }
```
NicEastvillage, line 252: Pretty sure this could just be
```rust
Some(LinearExpression { symbol: symbol_of_id, constant: number, operation: operator })
```
There seem to be no difference between the match arms, right?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,515 @@
...
+                self.cache.insert((expressions.clone(), target.state()), expr_distance as i32);
+                return Some(expr_distance / expressions.len() as f32);
+            }
+        }
+        None
+    }
+
+    fn get_linear_expressions_from_atlvertex(&self, vertex: &ATLVertex) -> Option<Vec<LinearExpression>> {
+        // get propositions from the formula in the vertex
+        let propositions = vertex.formula().get_propositions_recursively();
```
NicEastvillage, line 155: This does not seem correct. This way you do not consider the relation between the propositions/labels. E.g. you don't know if it is `A & B` or `A | B` or `!A & B`.

NicEastvillage (changes requested):
I would like you to go back to the MVP where we just consider formula for the form `x > 5`, such that we can make that work and merge it. After that we can extend to more advanced formula.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
MathiasMehl: fixed in 86f10b70fa28dd237a42c4d12482cdf6672b737c

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
MathiasMehl: Correct, fixed in d0e86a698edd1263f4f976d5e6a814c72d4ccb3f and e9d7146633c31d1378478c22cf39fad7c1983389

MathiasMehl:
> 
> 
> I would like you to go back to the MVP where we just consider formula for the form `x > 5`, such that we can make that work and merge it. After that we can extend to more advanced formula.

I believe this MVP is working now

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+/// based on distance from the vertex to a region that borders the line between true/false in the formula.
+pub struct LinearOptimizeSearch {
+    /// Priority based on distance, lowest distance highest priority
+    queue: PriorityQueue<Edge<AtlVertex>, i32>,
+    game: IntermediateLcgs,
+    /// Maps the hash of a Phi and usize of state, to a result distance
+    result_cache: HashMap<(Arc<Phi>, usize), i32>,
+    /// Maps usize from proposition, to a hashmap mapping symbols to the ranges they need to be within,
+    /// to satisfy the proposition
+    proposition_cache: HashMap<usize, HashMap<SymbolIdentifier, Ranges>>,
```
NicEastvillage, line 42: We have typealiases for `usize` called `State` and `Proposition` which you might want to use.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+    True,
+    False,
+}
+
+/// Holds extracted linear expressions from the formula in AtlVertex
+#[derive(Eq, PartialEq, Hash, Clone, Debug)]
+struct LinearExpression {
+    pub symbol: SymbolIdentifier,
+    pub constant: i32,
+    pub operation: BinaryOpKind,
```
NicEastvillage, line 86: Replace BinaryOpKind with a new enum or use ComparisonOp, since you know that it will never be addition/subtraction/division/etc anyway. Only comparisons.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+                    if let Some(state_of_symbol) = state.0.get(symbol) {
+                        let res = self.distance_to_range_bound(range, state_of_symbol);
+                        cumulative_distance = cumulative_distance + res;
+                    }
+                }
+
+                Some(cumulative_distance)
+            }
+            RangedPhi::True => Some(0),
+            RangedPhi::False => None,
```
NicEastvillage, line 488: This should be `Some(0)`. Otherwise `x < 5 || false` will also get a `None` distance.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+
+/// Used when mapping a Phi to a ranged variant,
+/// where the propositions have been replaced by hashmaps mapping symbols to Ranges
+pub enum RangedPhi {
+    /// Either or should hold
+    Or(Box<RangedPhi>, Box<RangedPhi>),
+    /// Both should hold
+    And(Box<RangedPhi>, Box<RangedPhi>),
+    /// Mapping symbols to ranges
+    Proposition(HashMap<SymbolIdentifier, Ranges>),
```
NicEastvillage, line 76: There is only every one element in this map, because we are only considering Propositions of the form of comparisons, like `x < 5`, right?

NicEastvillage (changes requested):
This is hard to get your head around this, so here is how I understand it:
* You assume propositions are simple comparison expressions with one identifier (eg. `x < 5`) and map each to the range, where the expression is true, if identifier is in that range.
* You map Phi to RangedPhi, which is a structure where each proposition has been replace with the range(s?) that satisfies the given comparison. You then visit this structure to calculate the min/max/summed distance from the state to the ranges.
* Stuff is cached.

And it works.

It's definitely overkill to you the LP library in the current state, but keeping as is it might make it easier to transitions to bigger linear expressions.

There are some minor things I would like you to look at. Most are tidiness. Clippy also complains a bit.

MathiasMehl:
> 
> 
> This is hard to get your head around this, so here is how I understand it:
> 
>     * You assume propositions are simple comparison expressions with one identifier (eg. `x < 5`) and map each to the range, where the expression is true, if identifier is in that range.
> 
>     * You map Phi to RangedPhi, which is a structure where each proposition has been replace with the range(s?) that satisfies the given comparison. You then visit this structure to calculate the min/max/summed distance from the state to the ranges.
> 
>     * Stuff is cached.
> 
> 
> And it works.
> 
> It's definitely overkill to you the LP library in the current state, but keeping as is it might make it easier to transitions to bigger linear expressions.
> 
> There are some minor things I would like you to look at. Most are tidiness. Clippy also complains a bit.

Yes, this is the idea. Should I add more comments/better explanations to explain the idea?
The idea came from trying to limit the number of times we use linear programming, as it has been pointed out to us, that the linear programming part could introduce a lot of overhead. So I figured if we could save the result from the linear programming solution somehow and reuse that, we could eliminate some of the overhead in this heuristic.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+        }
+    }
+}
+
+/// Holds the range for which a symbol should be within, to satisfy a proposition
+#[derive(Clone)]
+pub enum Ranges {
+    /// Must be the case that the state of the symbol is inside this range to satisfy the propositon (inclusive)
+    Range(i32, i32),
+}
```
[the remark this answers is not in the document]
MathiasMehl: Good point. There were at one point more variants in this enum, that why I have it instead of something else. I'll replace it

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+        linearexpression: &LinearExpression,
+        range_of_var: (f64, f64),
+        direction: OptimizationDirection,
+    ) -> Option<i32> {
+        let mut problem = Problem::new(direction);
+        let x = problem.add_var(1.0, (range_of_var.0, range_of_var.1));
+        match linearexpression.operation {
+            Addition => {
+                return None;
+            }
```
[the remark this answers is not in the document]
MathiasMehl: fixed in 94b632e

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
MathiasMehl: This case should be handled in `visit_ranged_phi` in the `Or` case.

If either of the lhs or rhs is None, the other is returned (given that it is some), else the min of the both is returned. 
```
RangedPhi::Or(lhs, rhs) => {
                if let Some(lhs_distance) = self.visit_ranged_phi(lhs, state) {
                    return if let Some(rhs_distance) = self.visit_ranged_phi(rhs, state) {
                        Some(Ord::min(lhs_distance, rhs_distance))
                    } else {
                        Some(lhs_distance)
                    };
                } else {
                    if let Some(rhs_distance) = self.visit_ranged_phi(rhs, state) {
                        return Some(rhs_distance);
                    }
                }
                None
            }
```

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
MathiasMehl: Yes

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -0,0 +1,505 @@
...
+                    let avg_distance = distances.iter().sum::<f32>() / distances.len() as f32;
+                    Some(avg_distance)
+                };
+            }
+            // Same procedure for negation edges as for hyper, just no for loop for all targets, as we only have one target
+            Edge::Negation(edge) => {
+                if let Some(distance) = self.get_distance_in_atl_vertex(&edge.target) {
+                    Some(distance as f32);
+                }
+                None
```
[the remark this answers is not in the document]
MathiasMehl: Like so? 2151476

I changed the return type of the function to i32 to avoid some conversions. It is at one point converted to f32 to find avg

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
NicEastvillage: Ah. Cool. Missed that. So None actually means undefined, and that makes sense.

NicEastvillage:
The amount of comments is fine. Just had to write it down with my own words. Talk to me when you are ready to start with bigger linear expressions.

falkecarlsen (changes requested):
I've read linear_optimize.rs but don't understand it enough to give constructive criticism. I've put some formatting and unnecessary change comments though.

atl-checker/src/game_structure/lcgs/ir/intermediate.rs
```diff
@@ -603,13 +610,10 @@ mod test {
         const max_health = 100;
         player alice = gamer;
         player bob = gamer;
-
```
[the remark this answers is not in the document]
falkecarlsen: Fixed

falkecarlsen (approved):

PR #151, opened 2021-05-06
Add graph feature to cli

NicEastvillage:
We lost the graph subcommand when breaking up the workspace. Now it is back!


falkecarlsen (approved):
Make graphing cool again :sunglasses:

PR #153, opened 2021-05-06
[P8-114] Add option to CLI to find partial strategy

NicEastvillage:
This PR adds the `--find-strategy <path>` or simply `-g <path>` option which will compute a partial game strategy that proves the result of the query. The partial game strategy is saved to the given path in a somewhat nice format.

Depends on #148 

Currently, computation of proofs only works when no proof is needed or when the formula is a true enforce property (except invariants).


cli/src/main.rs
```diff
@@ -141,18 +148,48 @@ fn main_inner() -> Result<(), String> {
...
+                            }
+                            SpecificationProof::NoStrategyNeeded => {
+                                println!("No game strategy was computed since a strategy is not needed to prove the given query.")
+                            }
+                        },
+                        Err(err) => {
+                            println!("Game strategy was not computed due to error: {}", err)
+                        }
+                    }
+                }
```
falkecarlsen, line 190: At some point, this stdout flow needs to match current main on handling quiet input

falkecarlsen (approved):

MathiasMehl (approved):

lars4509 (approved):

PR #158, opened 2021-05-12
[P8-77] Global Algorithm

lars4509:
Implemented the global algorithm, by reapplying `F_i` until the assignments does not change anymore. In order to do this correctly i have to discover the whole tree first, which is done by a depth first search, ignoring vertices already discovered.


atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+{
+    pub fn new(edg: G, v0: V) -> Self {
+        Self {
+            edg,
+            /// The root vertex, which we discover the graph from
+            v0,
+            /// The assignment for all vertices, we use this to keep track  of the current assignment
+            assignment: HashMap::<V, bool>::new(),
+            /// The distance of a vertex is represented as the index number of which set the vertex is located in
+            dist: VecDeque::<HashSet<V>>::new(),
```
NicEastvillage, line 30: These comments are typically put in the struct declaration.

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+        &mut self,
+        edge: Edge<V>,
+        queue: &mut VecDeque<(Edge<V>, u32)>,
+        dist: u32,
+    ) {
+        match edge {
+            Edge::Hyper(e) => {
+                for target in e.targets {
+                    if self.assignment.get(&target).is_some() {
+                        break;
```
NicEastvillage, line 83: This should be `continue` right?

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+    v0: V,
+    /// The assignment for all vertices, we use this to keep track  of the current assignment
+    assignment: HashMap<V, bool>,
+    /// The distance of a vertex is represented as the index number of which set the vertex is located in
+    dist: VecDeque<HashSet<V>>,
+}
+
+impl<
+        G: ExtendedDependencyGraph<V> + Send + Sync + Clone + Debug + 'static,
+        V: Vertex + Send + Sync + 'static,
```
NicEastvillage, line 22: I believe `Send + Sync + Clone + Debug + 'static` are all unnecessary for the global algorithm since it is not distributed.

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// Firing the global algorithm, which simply reapplying the F_i function explained in the paper
+    /// until the assignment does not change anymore. Lastly the assignment of the root, v0, is returned
+    pub fn run(&mut self) -> bool {
+        self.initialize();
+
+        let mut components = self.get_components();
+
+        while !components.is_empty() {
+            let component = components.pop_front().unwrap();
+            while self.f(component.clone(), self.assignment.clone()) {}
```
NicEastvillage, line 43: You are cloning the component on every iteration !? Use more refereces please 😅

NicEastvillage: You can also avoid cloning the assignments, by having two of them and alternating between them.

NicEastvillage (changes requested):

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+    fn update_assignment(&mut self, v: V, new_ass: bool) -> bool {
+        self.assignment
+            .get_mut(&v)
+            .map(|ass| {
+                if new_ass == *ass {
+                    false
+                } else {
+                    *ass = new_ass;
+                    true
+                }
```
NicEastvillage, line 193: This is confusing to read. How about
```
let old = *ass;
*ass = new_ass;
new_ass != old
```

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+        let mut components = VecDeque::<HashSet<V>>::new();
+        let mut component = HashSet::<V>::new();
+
+        while !self.dist.is_empty() {
+            for v in self.dist.pop_back().unwrap() {
+                component.insert(v);
+            }
+            components.push_back(component.clone());
+        }
+        components
```
NicEastvillage, line 214: Looks like my biggest comment got yeeted... Here we go again:
If I read this correctly, component C_0 is a superset of C_1 which is a superset of C_2 etc? This matches the math, but it is actually unnecessary. Once the assignments of configurations of C_k has been found, they are never going to change. So if we want better performance, the components can be mutually exclusive.

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+            false
+        } else {
+            let mut final_ass = true;
+
+            if *ass_from_earlier.get(&edge.target).unwrap() {
+                final_ass = false;
+            }
+
+            self.update_assignment(edge.source, final_ass)
+        }
```
NicEastvillage, line 178: Pretty use the following will do the job
```rust
let new_ass = !*self.assignment.get(&edge.target).unwrap();
self.update_assignment(edge.source, new_ass)
```

lars4509: When implementing this, i have troubles that some cases enter a infinite loop.

lars4509: Nevermind, i think i misunderstood you comment.

NicEastvillage: Yeah, the whole if-statement can be replace with the given fragment.

lars4509: I have played around with it, and i think we need the first check, since we are always raising the value of vertices. If we do not check the source then we can risk to lower it, depending of the target value.

lars4509: But the else branch can be replaced with the suggestion

NicEastvillage (approved):

cli/src/main.rs
```diff
@@ -262,6 +263,96 @@ fn main_inner() -> Result<(), String> {
...
+                        state: graph.game_structure.initial_state_index(),
+                        formula: arc,
+                    };
+                    let result = GlobalAlgorithm::new(graph, v0).run();
+                    println!("Result: {}", result);
+                    Ok(())
+                },
+            )??
+        }
+        ("global", Some(global_args)) => {
```
MathiasMehl, line 311: Am I seeing it wrong, or are there two cases for global?

atl-checker/src/algorithms/global/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+    pub fn new(edg: G, v0: V) -> Self {
+        Self {
+            edg,
+            v0,
+            assignment: HashMap::<V, bool>::new(),
+            dist: VecDeque::<HashSet<V>>::new(),
+        }
+    }
+
+    /// Firing the global algorithm, which simply reapplying the F_i function explained in the paper
```
MathiasMehl, line 26:
```suggestion
    /// Firing the global algorithm, which simply reapplies the F_i function explained in the paper
```

cli/src/main.rs
lars4509: For some reason there are, gonna fix that right away

MathiasMehl (approved):

PR #160, opened 2021-05-12
[P8-97] Linear constraint extraction for linear heuristic search

NicEastvillage:
Adds linear constraint extractor, allowing us to use linear heuristic for more than just simple comparisons.


atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
```diff
@@ -0,0 +1,565 @@
...
+                BinaryOpKind::Division => {
+                    // We found a potential term.
+                    // A variable divided with a constant is a linear expression,
+                    // but constant divided by variable is not.
+                    if let (ExprKind::OwnedIdent(ident), ExprKind::Number(n)) =
+                        (&lhs.kind, &rhs.kind)
+                    {
+                        if let Identifier::Resolved { owner, name } = ident.as_ref() {
+                            let coefficient = self.terms.entry(owner.symbol_id(name)).or_default();
+                            *coefficient += (1 / n) * sign;
```
MathiasMehl, line 147: Wont the right side almost always become 0, as we are dealing with i32?

NicEastvillage: It will. In general, using floats in the linear heuristic will give us more precision. We are going to need the `float-ord` crate to give a total ordering to floats (otherwise we can't do `max(float_dist_1, float_dist_2)`)

MathiasMehl: Yes, I believe that I also ran into that at some point. How much of an impact do you think this will have on measuring the distance, using ints?

NicEastvillage: Small, but some. It makes it more rough than it already is.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_optimize.rs
```diff
@@ -139,322 +124,195 @@ impl LinearOptimizeSearch {
...
+            ExprKind::UnaryOp(UnaryOpKind::Not, sub_expr) => {
+                return self.map_expr_to_constraints(sub_expr);
             }
-            // If we need to satisfy both of the formulas, just return the largest distance found between the two
-            RangedPhi::And(lhs, rhs) => {
-                if let Some(lhs_distance) = self.visit_ranged_phi(lhs, state) {
-                    if let Some(rhs_distance) = self.visit_ranged_phi(rhs, state) {
-                        return Some(Ord::max(lhs_distance, rhs_distance));
+            ExprKind::BinaryOp(operator, lhs, rhs) => {
+                match operator {
```
MathiasMehl, line 233: We should be able to add more cases here, implication, ternary and so on can be converted to a disjunctive/conjunctive normal form

NicEastvillage: Feel free to add those changes. I am working on floats.

MathiasMehl: I'll look into it

MathiasMehl: #163

MathiasMehl (approved):

lars4509 (approved):

PR #162, opened 2021-05-13
Fix atl formulae

### LINKED ISSUE #121: Mexican Standoff - the formulas and models does not align
It seems like the players in mexican standoff is 0-indexed and the in the formulas the players are 1-indexed, which leads to an error when running the models and formulas.

MathiasMehl:
I found a few errors in the ¿indices? used. Skimmed through all formulae, both json/atl and this was all that I found. Did not check if indices of propositions in the json versions were correct. 

fixed #121


GeneralDenmark (approved):
Good catch!

lars4509 (approved):
I appreciate you took time to fix this, i believe there migh be a bug in the "generator".

PR #163, opened 2021-05-13
More comparisons

MathiasMehl:
Adds cases for ternary, implication, number and owned ident


NicEastvillage (commented):
> we don't care about no

Not yet. But the linear programs will.

NicEastvillage (approved):

PR #164, opened 2021-05-13
[P8-111] Actual linear programming search

NicEastvillage:
- Adds linear programming search which priotises edges based on the source's minimal distance to satisfy the formula.
- Since linear programming also uses `LinearConstrainedPhi`, the extraction of `LinearConstrainedPhi` is now handled by `ConstrainedPhiMapper`.

Depends on #160

### DIFF
--- a/atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
+++ b/atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
@@ -1,0 +1,210 @@
+use crate::algorithms::certain_zero::search_strategy::linear_constraints::{
+    ComparisonOp, LinearConstraint, LinearConstraintExtractor,
+};
+use crate::atl::Phi;
+use crate::game_structure::lcgs::ast::{
+    BinaryOpKind, DeclKind, Expr, ExprKind, Identifier, UnaryOpKind,
+};
+use crate::game_structure::lcgs::ir::intermediate::IntermediateLcgs;
+use crate::game_structure::lcgs::ir::symbol_table::SymbolIdentifier;
+use crate::game_structure::Proposition;
+use std::borrow::{Borrow, BorrowMut};
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+/// A structure describing the relation between linear constraints in a ATL formula
+#[derive(Clone)]
+pub enum LinearConstrainedPhi {
+    /// Either or should hold
+    Or(Box<LinearConstrainedPhi>, Box<LinearConstrainedPhi>),
+    /// Both should hold
+    And(Box<LinearConstrainedPhi>, Box<LinearConstrainedPhi>),
+    /// Mapping symbols to ranges
+    Constraint(LinearConstraint),
+    True,
+    False,
+}
+
+pub struct ConstrainedPhiMapper {
+    /// Cached LinearConstrainedPhi of propositions
+    proposition_cache: RefCell<HashMap<Proposition, LinearConstrainedPhi>>,
+}
+
+impl ConstrainedPhiMapper {
+    pub fn new() -> ConstrainedPhiMapper {
+        ConstrainedPhiMapper {
+            proposition_cache: RefCell::new(HashMap::new()),
+        }
+    }
+
+    /// Map the given phi to a linear constrained phi. Propositions will be map based on the given
+    /// game and they will be cached, so the same game should be used every time.
+    pub fn map(&self, game: &IntermediateLcgs, phi: &Phi) -> LinearConstrainedPhi {
+        self.map_phi_to_constraints(game, phi)
+    }
+
+    /// Takes a phi and maps it to a LinearConstrainedPhi
+    fn map_phi_to_constraints(&self, game: &IntermediateLcgs, phi: &Phi) -> LinearConstrainedPhi {
+        match phi {
+            Phi::True => LinearConstrainedPhi::True,
+            Phi::False => LinearConstrainedPhi::False,
+            Phi::Proposition(proposition) => {
+                // If we get to a proposition, continue the mapping inside the proposition's condition
+                // The result may be cached
+                let maybe_constraint = self.proposition_cache.borrow().get(proposition).cloned();
+                maybe_constraint.unwrap_or_else(|| {
+                    // We have not mapped this proposition yet
+                    let decl = game.label_index_to_decl(*proposition);
+                    if let DeclKind::Label(label) = &decl.kind {
+                        let mapped_expr = self.map_expr_to_constraints(&label.condition);
+                        // Save result in cache
+                        self.proposition_cache
+                            .borrow_mut()
+                            .insert(*proposition, mapped_expr.clone());
+                        mapped_expr
+                    } else {
+                        panic!("Non-propositions symbol in ATL formula")
+                    }
+                })
+            }
+            Phi::Not(formula) => self.map_phi_to_constraints(game, formula),
+            Phi::Or(lhs, rhs) => {
+                let lhs_symbol_range_map = self.map_phi_to_constraints(game, lhs);
+                let rhs_symbol_range_map = self.map_phi_to_constraints(game, rhs);
+
+                LinearConstrainedPhi::Or(
+                    Box::from(lhs_symbol_range_map),
+                    Box::from(rhs_symbol_range_map),
+                )
+            }
+            Phi::And(lhs, rhs) => {
+                let lhs_symbol_range_map = self.map_phi_to_constraints(game, lhs);
+                let rhs_symbol_range_map = self.map_phi_to_constraints(game, rhs);
+
+                LinearConstrainedPhi::And(
+                    Box::from(lhs_symbol_range_map),
+                    Box::from(rhs_symbol_range_map),
+                )
+            }
+            Phi::DespiteNext { formula, .. } => self.map_phi_to_constraints(game, formula),
+            Phi::EnforceNext { formula, .. } => self.map_phi_to_constraints(game, formula),
+            Phi::DespiteUntil { pre, until, .. } => {
+                let pre_symbol_range_map = self.map_phi_to_constraints(game, pre);
+                let until_symbol_range_map = self.map_phi_to_constraints(game, until);
+
+                LinearConstrainedPhi::Or(
+                    Box::from(pre_symbol_range_map),
+                    Box::from(until_symbol_range_map),
+                )
+            }
+            Phi::EnforceUntil { pre, until, .. } => {
+                let pre_symbol_range_map = self.map_phi_to_constraints(game, pre);
+                let until_symbol_range_map = self.map_phi_to_constraints(game, until);
+
+                LinearConstrainedPhi::Or(
+                    Box::from(pre_symbol_range_map),
+                    Box::from(until_symbol_range_map),
+                )
+            }
+            Phi::DespiteEventually { formula, .. } => self.map_phi_to_constraints(game, formula),
+            Phi::EnforceEventually { formula, .. } => self.map_phi_to_constraints(game, formula),
+            Phi::DespiteInvariant { formula, .. } => self.map_phi_to_constraints(game, formula),
+            Phi::EnforceInvariant { formula, .. } => self.map_phi_to_constraints(game, formula),
+        }
+    }
+
+    /// Takes an expression and maps it to a LinearConstrainedPhi
+    fn map_expr_to_constraints(&self, expr: &Expr) -> LinearConstrainedPhi {
+        match &expr.kind {
+            ExprKind::UnaryOp(UnaryOpKind::Not, sub_expr) => {
+                return self.map_expr_to_constraints(sub_expr);
+            }
+            ExprKind::BinaryOp(operator, lhs, rhs) => {
+                match operator {
+                    BinaryOpKind::Equality
+                    | BinaryOpKind::GreaterThan
+                    | BinaryOpKind::GreaterOrEqual
+                    | BinaryOpKind::LessThan
+                    | BinaryOpKind::LessOrEqual => {
+                        let lin_expr = LinearConstraintExtractor::extract(expr);
+                        return if let Some(lin_expr) = lin_expr {
+                            LinearConstrainedPhi::Constraint(lin_expr)
+                        } else {
+                            // Not linear
+                            LinearConstrainedPhi::True
+                        };
+                    }
+                    BinaryOpKind::And => {
+                        let lhs_con = self.map_expr_to_constraints(lhs);
+                        let rhs_con = self.map_expr_to_constraints(rhs);
+                        return LinearConstrainedPhi::And(Box::new(lhs_con), Box::new(rhs_con));
+                    }
+                    BinaryOpKind::Or => {
+                        let lhs_con = self.map_expr_to_constraints(lhs);
+                        let rhs_con = self.map_expr_to_constraints(rhs);
+                        return LinearConstrainedPhi::Or(Box::new(lhs_con), Box::new(rhs_con));
+                    }
+                    // P -> Q == not P v Q, we don't care about not, so becomes P v Q
+                    BinaryOpKind::Implication => {
+                        let lhs_con = self.map_expr_to_constraints(lhs);
+                        let rhs_con = self.map_expr_to_constraints(rhs);
+                        return LinearConstrainedPhi::Or(Box::new(lhs_con), Box::new(rhs_con));
+                    }
+                    _ => {}
+                }
+            }
+            ExprKind::Number(n) => {
+                return if *n == 0 {
+                    LinearConstrainedPhi::False
+                } else {
+                    LinearConstrainedPhi::True
+                }
+            }
+            // https://en.wikipedia.org/wiki/Conditioned_disjunction
+            // Q ? P : R == (Q -> P) and (not Q -> R) == (Q and P) or (not Q and R)
+            // we don't care about not, so becomes (Q and P) or (Q and R), and can be written as
+            // (Q and (P or R))
+            ExprKind::TernaryIf(q, p, r) => {
+                let q = self.map_expr_to_constraints(q);
+                let p = self.map_expr_to_constraints(p);
+                let r = self.map_expr_to_constraints(r);
+                return LinearConstrainedPhi::And(
+                    Box::new(q),
+                    Box::from(LinearConstrainedPhi::Or(Box::new(p), Box::new(r))),
+                );
+            }
+            // Some other expression can be converted to a comparisons since everything != 0 is true
+            ExprKind::OwnedIdent(ident) => {
+                let mut terms_hashmap = HashMap::new();
+                if let Identifier::Resolved { owner, name } = ident.as_ref() {
+                    terms_hashmap.insert(owner.symbol_id(name), 1.0);
+                }
+
+                let less_constraint = LinearConstraint {
+                    terms: terms_hashmap.clone(),
+                    constant: 0.0,
+                    comparison: ComparisonOp::Less,
+                    coefficient_norm: 1.0,
+                };
+
+                let greater_constraint = LinearConstraint {
+                    terms: terms_hashmap,
+                    constant: 0.0,
+                    comparison: ComparisonOp::Greater,
+                    coefficient_norm: 1.0,
+                };
+
+                return LinearConstrainedPhi::Or(
+                    Box::new(LinearConstrainedPhi::Constraint(less_constraint)),
+                    Box::new(LinearConstrainedPhi::Constraint(greater_constraint)),
+                );
+            }
+
+            _ => {}
+        }
+
+        // Not linear
+        LinearConstrainedPhi::True
+    }
+}
--- a/atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
+++ b/atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
@@ -26,6 +26,26 @@ pub enum LinearConstrainedPhi {
     False,
 }
 
+impl LinearConstrainedPhi {
+    /// Returns the negations of this constrained phi
+    pub fn negated(&self) -> Self {
+        match self {
+            LinearConstrainedPhi::Or(lhs, rhs) => {
+                LinearConstrainedPhi::And(Box::new(lhs.negate()), Box::new(rhs.negate()))
+            }
+            LinearConstrainedPhi::And(lhs, rhs) => {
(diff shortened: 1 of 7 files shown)


atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl ConstrainedPhiMapper {
+    pub fn new() -> ConstrainedPhiMapper {
+        ConstrainedPhiMapper {
+            proposition_cache: RefCell::new(HashMap::new()),
+        }
+    }
+
+    /// Map the given phi to a linear constrained phi. Propositions will be map based on the given
```
MathiasMehl, line 41:
```suggestion
    /// Map the given phi to a linear constrained phi. Propositions will be mapped based on the given
```

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,253 @@
...
+                })
+            }
+            Phi::Not(formula) => self.map_phi_to_constraints(game, formula, !negated),
+            Phi::Or(lhs, rhs) => {
+                let lhs_lcp = self.map_phi_to_constraints(game, lhs, negated);
+                let rhs_lcp = self.map_phi_to_constraints(game, rhs, negated);
+                if !negated {
+                    LinearConstrainedPhi::Or(Box::from(lhs_lcp), Box::from(rhs_lcp))
+                } else {
+                    LinearConstrainedPhi::And(Box::from(lhs_lcp), Box::from(rhs_lcp))
```
MathiasMehl, line 96: DeMorgan?

NicEastvillage: Yes. Also notice that `lhs_lcp` and `rhs_lcp` were negated if `negated` was true.

MathiasMehl: Yes, I was about to make a comment about they should be negated in the new `or` and `and`, until I noticed this fact

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,253 @@
...
+            // https://en.wikipedia.org/wiki/Conditioned_disjunction
+            // Q ? P : R == (Q -> P) and (not Q -> R) == (Q and P) or (not Q and R)
+            ExprKind::TernaryIf(q, p, r) => {
+                let q = self.map_expr_to_constraints(q, negated);
+                let p = self.map_expr_to_constraints(p, negated);
+                let r = self.map_expr_to_constraints(r, negated);
+                return if !negated {
+                    LinearConstrainedPhi::Or(
+                        Box::new(LinearConstrainedPhi::And(Box::new(q.clone()), Box::new(p))),
+                        Box::from(LinearConstrainedPhi::And(Box::new(q), Box::new(r))),
```
MathiasMehl, line 215: Should'nt q be negated in the second `And` inside the `Or`?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,253 @@
...
+                    )
+                } else {
+                    LinearConstrainedPhi::And(
+                        Box::new(LinearConstrainedPhi::Or(Box::new(q.clone()), Box::new(p))),
```
MathiasMehl, line 219: And in this case, wont it become `(not q or not p) and (not q or not r)` ?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
MathiasMehl: image

The top table is what the ternary truth table is, and bottom is what I understand it is being translated to. Am I understanding the code wrongly?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// Proposition that is not linear. We can't use these in the linear methods, but it is not
+    /// definitively true or false either
+    NonLinear,
+    True,
+    False,
+}
+
+pub struct ConstrainedPhiMapper {
+    /// Cached LinearConstrainedPhi of propositions. The bool is true for negated propositions.
+    proposition_cache: RefCell<HashMap<(Proposition, bool), LinearConstrainedPhi>>,
```
MathiasMehl, line 31: Instead of passing Game around in the mapping function ,couldnt it just be part of the struct?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
```diff
@@ -0,0 +1,596 @@
...
+            ComparisonOp::GreaterOrEq => write!(f, ">="),
+        }
+    }
+}
+
+impl From<ComparisonOp> for minilp::ComparisonOp {
+    fn from(op: ComparisonOp) -> Self {
+        match op {
+            ComparisonOp::Equal | ComparisonOp::NotEqual => minilp::ComparisonOp::Eq,
+            ComparisonOp::Less | ComparisonOp::LessOrEq => minilp::ComparisonOp::Le,
```
MathiasMehl, line 47: Wont this become a problem for the linear program, that the greater/less comparisons are squished into one? Or maybe this is handled in the linear programming code by subtracting/adding 1? (have not read that part yet)

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/// Search strategy using ideas from linear programming to order next vertices,
+/// based on distance from the vertex to a region that borders the line between true/false in the formula.
+pub struct LinearProgrammingSearch {
+    /// Priority based on distance, lowest distance highest priority
+    queue: PriorityQueue<Edge<AtlVertex>, i32>,
+    game: IntermediateLcgs,
+    phi_mapper: ConstrainedPhiMapper,
+    /// Maps the hash of a Phi and usize of state, to a result state and distance
```
MathiasMehl, line 38: Is this correct?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+        // So we iterate through them and find the best result
+        let mut best: Option<(State, i32)> = None;
+        for constraints in all_variants(constrained_phi) {
+            // We keep track of the symbols/variables encountered
+            let mut symbol_vars: HashMap<SymbolIdentifier, Variable> = HashMap::new();
+            let mut problem = Problem::new(OptimizationDirection::Minimize);
+
+            for constraint in &constraints {
+                if constraint.comparison == ComparisonOp::NotEqual {
+                    // We skip not-equal constraints since the minilp cannot handle them.
```
MathiasMehl, line 129: Would we not be able to make a NotEqual into and `Or` with two cases, `less than` and `greater than`?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+                    sol_state.insert(symbol, v);
+                }
+
+                if let Some((_, old_best_dist)) = best {
+                    if old_best_dist > man_dist {
+                        // We found a better solution
+                        best = Some((State(sol_state), man_dist));
+                    }
+                } else {
+                    // We our first possible a solution
```
MathiasMehl, line 175: You a word here

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+                if constraint.comparison == ComparisonOp::NotEqual {
+                    // We skip not-equal constraints since the minilp cannot handle them.
+                    // They don't restrict the solution space in a meaning way anyway.
+                    continue;
+                }
+                // Build constraint
+                let mut lin_expr = LinearExpr::empty();
+                // The contraint is offset by the given state. This is reflected by adding
+                // coefficient*value_of_symbol_in_state to the constant
+                let mut offset = -constraint.constant;
```
MathiasMehl, line 137: I'm not quite sure whats going on here

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+/// Returns all variants of linear problems that can be created from the given LinearConstrainedPhi.
+/// Each problem consists of a list of linear constraints.
+fn all_variants(phi: &LinearConstrainedPhi) -> Vec<Vec<LinearConstraint>> {
+    all_variants_rec(vec![vec![]], phi)
+}
+/// Recursive helper function to [all_variants].
+fn all_variants_rec(
```
MathiasMehl, line 191: Also, not quite sure I grasp exactly how this works. But essentially it becomes a vec of all possible combinations of the ways we could satisfy the phi?

MathiasMehl:
I think theres going to be a lot of overhead in the linear programming in this. Is there perhaps some way we can reuse some of the linear programmin results, in addition to the existing result cache. I could imagine that not all linear programming here would be unique, and could perhaps be cached. 

Also when mapping expressions, perhaps cache here, mapping expressions to constraints?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+                self.queue.push(edge, 0);
+            }
+        }
+    }
+}
+
+impl LinearProgrammingSearch {
+    /// Finds the distance of an edge using linear programming
+    fn get_distance_of_edge(&mut self, edge: &Edge<AtlVertex>) -> Option<i32> {
+        let source = edge.source();
```
MathiasMehl, line 84: Did you find that source had better performance than targets?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
```diff
@@ -0,0 +1,224 @@
...
+        constrained_phi: &LinearConstrainedPhi,
+        state: &State,
+    ) -> Option<(State, i32)> {
+        // The LinearConstrainedPhi contains multiple variants of the linear problem due to the ORs.
+        // So we iterate through them and find the best result
+        let mut best: Option<(State, i32)> = None;
+        for constraints in all_variants(constrained_phi) {
+            // We keep track of the symbols/variables encountered
+            let mut symbol_vars: HashMap<SymbolIdentifier, Variable> = HashMap::new();
+            let mut problem = Problem::new(OptimizationDirection::Minimize);
```
MathiasMehl, line 125: Also, why Minimize? (not because I disagree, just to hear your thoughts)

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
NicEastvillage: I tried given a lifetimed reference to the game in the structure, borrow from LinearProgrammingSearch, but Rust was not happy about it. Found a forum post that said references to parent struct are generally difficult and the best alternatives are 1) cloning the thing, or 2) passing the reference on every call, or 3) use a `Rc<RefCell<T>>` (this option doesn't work for us since the search strategy has to be `'static` which Rc isn't). So I those the second option since there will be too many clones otherwise.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
NicEastvillage: If you look in `linear_programming_search.rs:128-132` you will find that NotEqual's are actually skipped right now.
```
// We skip not-equal constraints since the minilp cannot handle them.
// They don't restrict the solution space in a meaning way anyway.
```
NotEqual's are true everywhere, except on the hyperplane (line), so they are basically true everywhere.
As you mention, we could simulate it with OR, but then we also risk doubling the amount of optimization problems.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: Technically, yes. But it is a copy-paste error too.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: Yes. So if p, q, r, s, t are constraints then `(p OR q) AND (r OR (s AND t))` becomes `[p+r, q+r, p+s+t, q+s+t]`. It is similar to generating all words in the context free grammar
```
S -> AB
A -> p | q
B -> r | st
```
Some reference method for doing this: https://math.stackexchange.com/questions/36403/generating-all-words-in-a-language-from-cfg

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: The new `2021-05-14_13-29-22-bench-lin-source-dist-bench-lin-ave-target-dist.tar.gz` benchmark results will show us

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: We want to find the closest satisifying state, aka minimum distance.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: This is where we take the state into account. First of all, note that minilp expects the constant on the RHS instead of LHS like our LinearConstraint struct, which is why the constant is negative.

Now if the constraint is `ax+by+cz<k` and we want to find the closest solution to the state `(x0, y0, z0)`, then we have to translate (move) the constraint by subtracting the state, such our the state becomes the origin. This becomes `a(x-x0)+b(y-y0)+c(z-z0)<k` => `ax-ax0+by-by0+cz-cz0<k` => `ax+by+cz<k+ax0+by0+cz0`.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
NicEastvillage: Good catch!

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
MathiasMehl: This comment was meant not about the Equal/NotEqual, but about the other two, Less/LessOrEq into just Le, and Greater/GreaterOrEq into Ge. We lose some precision in this, yes? x < 5 is not the same expression as x <= 5, but in the linear program wont these two expression become the same constraint?

NicEastvillage:
> I think theres going to be a lot of overhead in the linear programming in this.

I think so too, but in my preliminary tests the speed was on par with the other linear optimize search. We should avoid premature optimizations. Ideally we use our tests to determine if/where changes need to be made.

> Also when mapping expressions, perhaps cache here, mapping expressions to constraints?

We are already caching the constrained phi of propositions, which captures all the constraints on a good level. E.g. `<<A>> F (p & q)` will branch into `p` and `q` at some point, but it will never involve other propositions than `p` and `q`. So once we have cached those propositions (which we do on the first exploration), then we never look at expressions again.

> Also, I did not read much through the linear optimize, which I assume has not been touched much since last PR?

Correct. Just replaced the phi mapping to use the new ConstrainedPhiMapper.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
NicEastvillage: Just skimmed through the report and using source seems to be slightly better. I think the intuition for it is, that
1) In the worst case, you have to explore a big part of the graph anyway, so calculating the average of targets each time is just overhead.
2) In the best and average cases, where we terminate earlier, calculating the distance for the targets, is still overhead, but you are not going to visit many of those targets anyway.
When using the source, we have more edges with the same priority, but only calculate the distance for vertices that we actually visit.

This is speculation, though. Ideally it should be measured in tests.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
MathiasMehl: Visuals made by paint (not .net) gang

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
MathiasMehl: And we could add a comment about being able to handle the NotEqual, but we choose not to?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
MathiasMehl: Coolio

MathiasMehl: Thats a good point, maybe we can use that in another paper

MathiasMehl:
> 
> 
> > I think theres going to be a lot of overhead in the linear programming in this.
> 
> I think so too, but in my preliminary tests the speed was on par with the other linear optimize search. We should avoid premature optimizations. Ideally we use our tests to determine if/where changes need to be made.

True. I was just thinking about ways to reduce some overhead, and if we don't see much increase in speed with this, that and Mathias' supervisor suggestion could be next step

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
MathiasMehl: image

So like this? (circle is our state)

NicEastvillage: Yes. So ATL vertices with similar formula might have different solutions because the state is very different.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
NicEastvillage: > This comment was meant not about the Equal/NotEqual, but about the other two,

I see 🤔 Well, again, by adding an OR we risk doubling the amount of linear problems that need to be solved. And the difference between a `x < 5` and `x <= 5` solution is very small in terms of distance.

MathiasMehl: Thats true, but we could handle the cases for `<=` and `>=` by respectively adding/subtracting one to the constant. Such that `x <= 5` becomes a `x < 6` constraint, and `x >= 5` becomes a `x > 4` constraint?

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
```diff
@@ -0,0 +1,274 @@
...
+                let q = self.map_expr_to_constraints(q, negated);
+                let not_q = q.negated();
+                let p = self.map_expr_to_constraints(p, negated);
+                let r = self.map_expr_to_constraints(r, negated);
+                return if !negated {
+                    LinearConstrainedPhi::Or(
+                        Box::new(LinearConstrainedPhi::And(Box::new(q), Box::new(p))),
+                        Box::new(LinearConstrainedPhi::And(Box::new(not_q), Box::new(r))),
+                    )
+                } else {
```
MathiasMehl, line 238: I'm not sure that the ternary is correctly mapped in the  `else`  case. As I can see it, it becomes `Q ? P : R` == `(not q Or not p) and (not q Or not r)`

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constraints.rs
NicEastvillage: Hold up. Just realised something. `minilp::ComparisonOp::Le` is for less *or equal*, and `minilp::ComparisonOp::Ge` greater *or equal*.

MathiasMehl: Well that's rude of them

atl-checker/src/algorithms/certain_zero/search_strategy/linear_programming_search.rs
MathiasMehl: The reason I ask, is because I wondered myself what to do here. Because, if we want to measure distance to the "acceptance border", or the point on satisfying/not satisfying, right? In that case, 
image
If we minimize in this scenario, where state is at origin, green area is acceptance, then will we get a distance of 0, because the linear programming will try to minimize all variables in the solution. Same sort of problem in the maximize direction. It is only if we are "below" the constraint and maximize in this situation we get the distance to the constraint. 
And in the same way, it is only if we are "above" the constraint and minimize, we get the distance to the constraint. 
Or am I understanding it wrongly?

NicEastvillage: With linear programming we want to find the minimum distance to a state that satisifies the formula, not the distance to the acceptance border. The other heuristic we made searches towards acceptrance borders, which is why it is so unique.

atl-checker/src/algorithms/certain_zero/search_strategy/linear_constrained_phi.rs
NicEastvillage: If negated, then we want to produce `not (Q and P) or (not Q and R)` => `(not (Q and P)) and (not (not Q and R))` => `(not Q or not P) and (Q or not R)`. And all terms are negated since negated was passed to it, so ultimately we want `(Q or P) and (not Q or R)`. So looks like I should swap the `q` and `not_q`. Heh.

MathiasMehl (approved):

lars4509 (approved):
I have hard time grasping the concept but it looks good.

PR #168, opened 2021-05-18
Add partial game strategy computation feature

NicEastvillage:
This PR finishes the development of computation of partial game strategies. It adds the `--find-strategy <path>` or `-g <path>` option, that when set, will make the solver find the partial game strategy that proves the (dis)satisfaction of the property in question and saves the strategy to the given path.

Not all types of strategies can be found yet. This initial version supports all enforce properties (except invariant) that are true.

The interesting thing in this PR is how it reworks main to incorporate all features.


atl-checker/src/algorithms/certain_zero/game_strategy/mod.rs
```diff
@@ -0,0 +1 @@
+
```
asger-weirsoee, line 1: dette ser ikke helt rigtigt ud

asger-weirsoee (approved):
With the exception of a few Clippy warnings introduced in this PR in the following files:

- atl-checker/src/algorithms/certain_zero/mod.rs 
- atl-checker/src/edg/atledg/format.rs
- cli/src/solver.rs 

The model checker still runs and the code is of high quality, hence the Approve

MathiasMehl (approved):
A few clippy warnings with unreachable code and some redundant return statements I believe, but otherwise looks good. Oh, I see Asger also commented one these, didn't see that until now

falkecarlsen (approved):

PR #170, opened 2021-05-19
Fix global subcommand stdout and return type

falkecarlsen:


MathiasMehl:
Looks good, but I think the #168 might create some merge conflict with this

falkecarlsen:
Fix for the fix incoming

MathiasMehl (approved):
Looks good, #168 might create some merge conflicts, heads-up.

falkecarlsen:
> Looks good, #168 might create some merge conflicts, heads-up.

You're right, it refactors it entirely. Merge is fixed

cli/src/main.rs
```diff
@@ -189,16 +189,32 @@ fn main_inner() -> Result<(), String> {
...
+                if result {
+                    if !quiet_flag {
+                        println!("Model satisfies formula: {}", result);
+                    }
+                    std::process::exit(42);
+                } else {
+                    if !quiet_flag {
+                        println!("Model satisfies formula: {}", result);
+                    }
+                    std::process::exit(43);
```
NicEastvillage, line 203: Remove the duplicate code by moving the print outside the outer if.

cli/src/main.rs
```diff
@@ -541,7 +557,15 @@ fn parse_arguments() -> ArgMatches<'static> {
...
                 .add_formula_arg()
-                .add_formula_type_arg(),
+                .add_formula_type_arg().arg(
+                Arg::with_name("quiet")
+                    .short("q")
+                    .takes_value(false)
+                    .long("quiet")
+                    .help(
+                        "Suppress stdout and only return exitcode 42 for true result or 43 for false",
+                    ),
```
NicEastvillage, line 567: Since this is now applies to two subcommands it should be moved to `cli::args`

cli/src/main.rs
falkecarlsen: Fixed in e6e2663

cli/src/main.rs
falkecarlsen: Fixed in e6e2663

PR #177, opened 2021-06-07
Add new, short suite for quick CI turnaround

falkecarlsen:
Does what it says on the tin :tm: 

If paths of any declared bench is wrong, the `cargo test --benches` flow fails, but we don't waste time on running each bench once for every CI check.

In future for real benchmarks, the user has to disable this suite - but it's described in comment directly at criterion_main function call.

### DIFF
--- a/benches/benchmark_solver.rs
+++ b/benches/benchmark_solver.rs
@@ -2547,13 +2547,16 @@ criterion_group!(
     rand_5p_5m_3000d_state_enforce_until_1,
     rand_5p_5m_3000d_state_enforce_until_2,
 );
+// tiny suite for shorter github CI turnaround, check still fails if any path in any declared bench is wrong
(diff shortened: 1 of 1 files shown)


falkecarlsen:
CI for benches now takes ~2:45 which is 13 seconds less than ordinary test-suite.

asger-weirsoee (approved):
5% faster. I mean it is something 🤷‍♂️

falkecarlsen:
> 5% faster. I mean it is something man_shrugging

No, you misunderstand - last run on main took 36 minutes: https://github.com/d702e20/CGAAL/runs/2710055687
So it's 92% faster now.

asger-weirsoee:
Ah, well then I misunderstood the  "~2:45 which is 13 seconds less than ordinary test-suite." then ~

NicEastvillage (approved):

PR #179, opened 2023-05-30
update emails on continuing authors and fix regex-bug in tracing-subs…

falkecarlsen:
…criber

### DIFF
--- a/atl-checker/Cargo.toml
+++ b/atl-checker/Cargo.toml
@@ -17,7 +17,7 @@ graph-printer = []
 use-counts = []
 
 [dev-dependencies]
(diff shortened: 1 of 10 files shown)


atl-checker/Cargo.toml
```diff
@@ -5,11 +5,11 @@ authors = [
     "d702e20 <d702e20@cs.aau.dk>",
     "d802f21 <d802f21@cs.aau.dk>",
```
NicEastvillage, line 6: Might as well remove these too.

NicEastvillage (approved):

PR #180, opened 2023-05-30
Evaluation

falkecarlsen:


NicEastvillage:
See #187 for fixing the failing test

falkecarlsen:
While this is not the method of evaluation that we end up using on the MCC, it still holds value for local testing.

NicEastvillage (approved):

PR #182, opened 2023-05-31
[clean] Leftover changes from master

NicEastvillage:
Turns out there is both a main and a master branch. Master has a single extra commit that gets rid of some warnings.


NicEastvillage:
Nvm, these changes are also on main. I'll delete master then

PR #184, opened 2023-05-31
Support for multiple threads using the global algorithm

lars4509:
I have implemented at simple solution for a multithreaded global algorithm. The idea is to have an producer-consumer setup, where the workers exhaust a queue while the master populate it with tasks.  Here is the overview:

As the single threaded version the whole EDG is explore beforehand and splitted up in components relative to the initial vertex, v0, by the master.

Then the master spawns n workers and starts populating the work-queue with the vertices of the current component. When a worker have calculated the value of a vertice it sends it back to the master - which updates its assignments and note if the updates is a change. The master requeues all the vertices in the component if there have been changes. 

When the queue is empy, all vertices in the current component have been processed and nothing have changed since last run. The vertices of the next combonent is queued, and the process is repeated.

When all components have been processed and nothing changes, all the workers is terminated and the master returns with the value of v0.

### DIFF
--- a/atl-checker/src/algorithms/global/com.rs
+++ b/atl-checker/src/algorithms/global/com.rs
@@ -1,56 +1,74 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap};
 use std::error::Error;
 use std::hash::Hash;
 use crossbeam_channel::{Receiver, Sender, TryRecvError};
 
 #[derive(Clone, Debug)]
 pub enum GMessage<V: Hash + Eq + PartialEq + Clone> {
-    Assignment {
-        assignment: HashMap<V,bool>,
+    Updates {
+        updates:HashMap<V, bool>,
+    },
+    Result {
+        task: V,
+        value: bool,
     },
     Terminate,
 }
 
 pub trait GBroker<V: Hash + Eq + PartialEq + Clone> {
-
-    fn return_changes(&self, assignments: HashSet<(V, bool)>);
-
+    fn send_terminate(&self);
+    fn send_result(&self, task: V ,value: bool);
     fn receive(&self) -> Result<Option<GMessage<V>>, Box<dyn Error>>;
-
     fn get_task(&self) -> Result<Option<V>, Box<dyn Error>>;
+    fn task_queue_is_empty(&self) -> bool;
+
 }
 
 pub trait GBrokerManager<V: Hash + Eq + PartialEq + Clone> {
-    fn send_assignments(&self, assignments: HashMap<V,bool>);
+    fn send_updates(&self, assignments: HashMap<V,bool>);
     fn queue_task(&self, task: V);
-    fn receive_changes(&self) -> Result<HashSet<(V, bool)>, Box<dyn Error>>;
+    fn receive(&self) -> Result<GMessage<V>, Box<dyn Error>>;
+    fn task_queue_is_empty(&self) -> bool;
     fn terminate(&self);
-
 }
 
 pub struct GChannelBrokerManager<V: Hash + Eq + PartialEq + Clone> {
     workers: Vec<Sender<GMessage<V>>>,
-    changes: Receiver<HashSet<(V, bool)>>,
-    task_queue: Sender<V>,
-
+    updates: Receiver<GMessage<V>>,
+    task_queue_producer: Sender<V>,
+    task_queue_consumer: Receiver<V>,
 }
 
 impl<V: Hash + Eq + PartialEq + Clone> GBrokerManager<V> for GChannelBrokerManager<V>{
-    fn send_assignments(&self, assignment: HashMap<V,bool>){
-        let msg = GMessage::Assignment {assignment};
+    fn send_updates(&self, updates: HashMap<V, bool>){
+        let msg = GMessage::Updates { updates };
         for worker in &self.workers{
             worker.send(msg.clone()).expect("Sending assignments to workers failed");
         }
     }
     fn queue_task(&self, task: V){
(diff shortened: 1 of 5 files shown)


atl-checker/src/algorithms/global/mod.rs
```diff
@@ -1,121 +1,45 @@
...
-    }
+pub trait GlobalAlgorithm<G: ExtendedDependencyGraph<V>, V: Vertex> {
+    fn edg(&self) -> &G;
+    fn edg_mut(&mut self) -> &mut G;
+    fn v0(&self) -> &V;
+    fn v0_mut(&mut self) -> &mut V;
+    fn assignment(&self) -> &HashMap<V, bool>;
+    fn assignment_mut(&mut self) -> &mut HashMap<V, bool>;
+    fn dist(&self) -> &VecDeque<HashSet<V>>;
+    fn dist_mut(&mut self) -> &mut VecDeque<HashSet<V>>;
```
NicEastvillage, line 17: Can you find a better navn for these?

atl-checker/src/algorithms/global/com.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.updates
+            .send(GMessage::Terminate)
+            .expect("Failed to send terminate message");
+    }
+
+    fn send_result(&self, task: V, value: bool) {
+        self.updates
+            .send(GMessage::Result { task, value })
+            .expect("Failed to send the assignments table to main thread");
+    }
```
NicEastvillage, line 99: I spot some copy-pasting for the local algorithm. Double check that the function names and error messages still make sense. For instance, here is says "assignments table" even though it is only a single assignment.

atl-checker/src/algorithms/global/multithread.rs
```diff
@@ -0,0 +1,484 @@
...
+    }
+    fn dist(&self) -> &VecDeque<HashSet<V>> {
+        panic!("A GWorker doesnt have a dist vecdeque")
+    }
+    fn dist_mut(&mut self) -> &mut VecDeque<HashSet<V>> {
+        panic!("A GWorker doesnt have a dist vecdeque")
+    }
+    fn run(&mut self) -> bool {
+        self.run()
+    }
```
NicEastvillage, line 49: This looks like recursion. But I assume it goes to the other definition of `run`. Heh

atl-checker/src/algorithms/global/multithread.rs
```diff
@@ -0,0 +1,484 @@
...
+                    if manager.task_queue_is_empty() && !changed_flag {
+                        break;
+                    }
+
+                    manager.send_updates(self.assignment().clone());
+                    for vertex in component {
+                        self.queue_task(vertex.clone(), &mut tasks, &manager);
+                    }
+                    changed_flag = false;
+                }
```
NicEastvillage, line 182: Is it important to keep track of which tasks are still queued? If we just care about if any task is still running, then you could use an counter instead of a hashmap. Increase the counter whenever you queue a task, decrease when it is complete. It is faster to increase/decrease/check-if-0 than it is to compute the hashes of vertices over and over.

cli/src/main.rs
```diff
@@ -219,7 +226,15 @@ fn main_inner() -> Result<(), String> {
...
                     };
-                    GlobalAlgorithm::new(graph, v0).run()
+
+                    if threads > 1 {
+                        MultithreadedGlobalAlgorithm::new(graph, threads, v0).run()
+                    } else if threads == 1 {
+                        SinglethreadedGlobalAlgorithm::new(graph, v0).run()
+                    } else {
+                        println!("The number of threads have to be positive");
+                        panic!("The number of threads have to be positive")
```
NicEastvillage, line 236: Don't panic. Produce error instead.

cli/src/main.rs
```diff
@@ -230,7 +245,15 @@ fn main_inner() -> Result<(), String> {
...
```
NicEastvillage, line 255: Don't panic. Produce error instead.

NicEastvillage (changes requested):
Nice work!

~~Clippy has some suggestions. Please fix those relevant for the additions in this PR.~~ You fixed these while I was reviewing. Nice.

A few more comments would be nice.

NicEastvillage:
The changes from #183, especially the `test_log`, affects this PR.

lars4509:
Right now, the test is only testing a single threaded version which is not optimal. Running the test with multiple workers result in the `test_multi_global_algorithm_negation_to_undecided_01()` failing sometimes. I am investigating it right now, but it is not obivious to me why this is the case.

falkecarlsen:
> Right now, the test is only testing a single threaded version which is not optimal. Running the test with multiple workers result in the `test_multi_global_algorithm_negation_to_undecided_01()` failing sometimes. I am investigating it right now, but it is not obivious to me why this is the case.
> 
> EDIT: It fails on multiple test just rarely. The problems seems to rely in the "changed" check.

Maybe use the CLion debugger to get out the assignment table for passing and failing tests, and comparing these would help investigate under which circumstance the algorithm fails. Afaik, Clion has a nice left->equal<-right comparison function.

NicEastvillage:
I am looking into it too. It surprises me that G gets the wrong assignment sometimes.

One hypothesis I had was: Maybe, after a component is requeued, a worker picks and performs a task before applying assignment updates from the master, resulting in a wrong answer to the task. However, a quick restructure, such that it picks a task, then updates, then perform the tasks, shows that there is more to it.

NicEastvillage (approved):

NicEastvillage:
See #187 for a fix to the failing test

PR #185, opened 2023-06-01
Fixed Clippy warnings

lars4509:
I have fixed most Clippy warnings but the ones in LPS since they will be fixed in #181.


falkecarlsen (approved):

NicEastvillage (approved):
Gotta make Clippy happy

PR #193, opened 2023-06-16
Add Linear Representative search strategy

NicEastvillage:
A search strategy suggested by Mathias CJ.

The LinearRepresentativeSearch is a search strategy that uses linear programming and the LCGS's representation of states in order to prioritize the edges during the search. However, it will only solve the linear programming problem once for the root formula, and use the solution states as representatives for the feasible regions of the formula. Edges are then prioritized based on their Manhattan distance to the representative states.

Pressumably, this works well for reachability queries in models where similar states are structurally close, but it will have a much smaller overhead than the LPS search.


NicEastvillage:
The CI error is fixed in #194

falkecarlsen (approved):
LGTM :shipit: 
Let's get this performance evaluated

PR #194, opened 2023-06-19
Fix CI failing to compile parser and remove lazy_static dependency

NicEastvillage:
The CI suddenly started to refuse compiling the parser. It can no longer infer the type of our list of reserved keywords. See https://github.com/d702e20/CGAAL/actions/runs/5308631315/jobs/9608318973?pr=193#step:4:164

image

This PR fixes that and removes the `lazy_static` dependency.


falkecarlsen (approved):
Nice :shipit

PR #196, opened 2023-06-26
Fix return code for CLI

falkecarlsen:
We always expect a non-zero return code to indicate errors, and a zero to indicate success. However, we had 42 for formula true, and 43 for formula false which is meme'y and wrong.


falkecarlsen:
all CI is failing, yet working locally :shipit:

PR #197, opened 2023-06-27
add threads argument to global alg cli parser

falkecarlsen:
Was missing from #184


NicEastvillage (approved):
Interesting how this was missed

falkecarlsen:
Another point we might as well write an issue for, is that it's dumb we have:
`atl-solver-cli solver -f formula.atl -m model.lcgs --search-strategy=bfs`
but not
`atl-solver-cli solver -f formula.atl -m model.lcgs --search-strategy=global`
instead we have to do
`atl-solver-cli global -f formula.atl -m model.lcgs`

This because both invocations are solving some formula on a model. Perhaps the --search-strategy arg should be rewritten to --method, to be more correct. Point still stands.

PR #201, opened 2023-06-30
Add prismlike robot coordination models

NicEastvillage:
Robot coordination models that matches https://github.com/d702e20/Prism-modeling/pull/16


falkecarlsen (approved):
Looks right. The ascii figures are really only correct for N=3 and 4 but whatever.

falkecarlsen (changes requested):
On second though, we should add these new models to the Criterion-testing harness for completeness.

NicEastvillage:
The bench will take a very long time to run, but ok. I have added RC4 and RC5 to the bench.

falkecarlsen (approved):

PR #202, opened 2023-07-03
Gossipping girls gone wild

lars4509:
Expanded current GG models with 5 and 6 players, and updated queries as well


benches/benchmark_solver.rs
```diff
@@ -423,86 +423,86 @@ bench_lcgs!(
 // gossiping girls
 bench_lcgs!(
     gg1_circular,
```
NicEastvillage, line 425: I renamed these to `ggN_circularQ` where `N` is the number of girls and `Q` is the query index. That way you can add `gg5_circularQ` to the bench as well.

NicEastvillage (changes requested):
Found this

falkecarlsen (approved):
:shipit:

benches/benchmark_solver.rs
```diff
@@ -2387,20 +2387,20 @@ criterion_group!(
...
-    gg6_total_threads,
-    gg7_total_threads, // gg_total benches takes avg 69s a piece (nice)
+    gg4_circular1_threads,
+    gg4_circular2_threads,
+    gg4_circular3_threads,
+    gg4_circular4_threads,
+    gg4_circular5_threads,
+    gg4_circular6_threads,
+    gg4_circular7_threads,
+    gg4_total1_threads,
```
NicEastvillage, line 2397: Also add `gg5_circ` to bench

NicEastvillage (changes requested):

benches/benchmark_solver.rs
NicEastvillage: Nvm

NicEastvillage (approved):

Issue #174, opened 2021-05-30
Wrong partial game strategy for robot_grid/exist_path_to_targets_with_no_crashes


NicEastvillage:
Computation of partial game strategies do not find the right result for `robot_grid/exist_path_to_targets_with_no_crashes_TRUE.atl` despite getting the right assignment.

Command:
```
atl-checker-cli.exe solver -m lcgs-examples/robot_grid/robot_grid_N3.lcgs -f lcgs-examples/robot_grid/exist_path_to_targets_with_no_crashes_TRUE.atl -g strategy.txt`
Checking the formula: <<r1,r2,r3,r4>> F (!(:global.any_crashes) & (r1.at_target & (r2.at_target & (r3.at_target & r4.at_target))))
Model satisfies formula: true
Proving game strategy was saved to strategy.txt
```
strategy.txt:
```
Partial strategy for the player(s): r1, r2, r3, r4
13752: r1.wait, r2.wait, r3.wait, r4.wait
```
Obviously that strategy won't make the robots arrive at their targets.

NicEastvillage:
Cause is likely #154

PR #207, opened 2023-07-04
Change invariant encoding from negated until to negated eventually

NicEastvillage:
While trying to find a solution to #174 I noticed the following:
Currently `G phi` is encoded as `!(true U !phi)`, which is correct. But an equivalent encoding is `!F !phi` (consult <https://en.wikipedia.org/wiki/Linear_temporal_logic> if needed, also note `F phi` is equivalent to `(true U phi`). And the `!F !phi` is slightly smaller than `!(true U !phi)` and will involve fewer edges and nodes. Hence, I propose changing the encoding of `G phi` to `!F !phi`. The change has been benchmarked and shown to be faster.

Benchmark setup: Using `Mexican_standoff_3p_3hp` with query `<<p1>> G p1.alive` (survive).

Benchmark results:
```
Benchmarking mexican_standoff_3p_3hp_lcgs_survive
Benchmarking mexican_standoff_3p_3hp_lcgs_survive: Warming up for 3.0000 s

Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 12.0s, or reduce sample count to 40.
Benchmarking mexican_standoff_3p_3hp_lcgs_survive: Collecting 100 samples in estimated 12.022 s (100 iterations)
Benchmarking mexican_standoff_3p_3hp_lcgs_survive: Analyzing
mexican_standoff_3p_3hp_lcgs_survive
                        time:   [118.38 ms 121.98 ms 125.61 ms]
                        change: [-11.382% -8.4251% -5.3900%] (p = 0.00 < 0.05)
                        Performance has improved.
```

image


falkecarlsen (approved):
We should compare the entire suite against main to see how much the total performance change is - but nice :shipit:

PR #208, opened 2023-07-05
Fix issues in strategy synthesis

### LINKED ISSUE #174: Wrong partial game strategy for robot_grid/exist_path_to_targets_with_no_crashes

NicEastvillage:
This PR reworks the strategy synthesis algorithm to resolve some bugs:

- [x] Fix positive cycles confusing the synthesis in enforce-until/eventually cases (Resolves #174)
- [x] Add synthesis of cases where despite-until vertices are assigned false (due to both sub formula being false)
- [x] Rewrite recursion to loop to avoid stack overflows
- [x] Test correctness of synthesized strategies (at least as much as possible for now)

Third change was needed since the algorithm would hit StackOverflow panics on any model of significant size. Hence, I had to rewrite the recursive algorithm to a work-list algorithm. At the same time I decided to split the true, false, and undecided assignment cases, which hopefully makes the algorithm(s) more readable.


cgaal-engine/src/algorithms/game_strategy/partial.rs
```diff
@@ -31,241 +32,265 @@ impl PartialStrategy {
...
-) -> PartialStrategy {
-    let mut move_to_pick = HashMap::new();
+) -> HashMap<State, PartialMove> {
+    debug_assert!(v0.formula().is_enforce());
+    debug_assert_eq!(assignments[v0], VertexAssignment::True);
 
-    compute_partial_strategy_rec(graph, v0, assignments, &mut move_to_pick);
+    if v0.formula().is_invariant() {
+        // EnforceInvariant has negation edge to DespiteEventually assigned undecided
+        let edges = graph.succ(v0);
```
larsbpf, line 48: These edges, doesn't seem to get used anywhere

cgaal-engine/src/algorithms/game_strategy/partial.rs
```diff
@@ -31,241 +32,265 @@ impl PartialStrategy {
...
-        move_to_pick,
+    if v0.formula().is_next() {
+        // Find first edge with all targets true
+        for edge in graph.annotated_succ(v0) {
+            let all_targets_true = edge
+                .targets()
+                .iter()
+                .all(|(target, _)| assignments[target].is_true());
+            if all_targets_true {
+                let mut move_to_pick = HashMap::<State, PartialMove>::new();
```
larsbpf, line 61: This variable name seems to only refer to a single move - but as I now understand is that is actually a map of states and partial moves. I suggest we rename it.

larsbpf (approved):
Looks great, one minor suggestion.

falkecarlsen (approved):

cgaal-engine/src/algorithms/game_strategy/partial.rs
NicEastvillage: I am not sure what you are arguing. Do you want to change it because it only has one entry? Or do you want to change it because it can hold more entries?

All the strategy synthesis functions in `partial.rs` return a map describing which partial move the coalition must pick in each state. In all the functions, the computed map is consistently called `move_to_pick` no matter the number of entries, since it describes which "move to pick" for each state.

falkecarlsen: It's nitpicky and could be either (actually a third)

> In all the functions, the computed map is consistently called move_to_pick no matter the number of entries, since it describes which "move to pick" for each state.

This gives rise to the naming `move_to_pick_for_state` or something equally verbose.

larsbopark: I understand that its consistently named `move_to_pick` but I suggest we find another name since it is ambigious in the way it can refer to a "single" move, and that is not the case, but instead it refer to a map. 

IF we choose to rename it ofcourse it should be renamed consistently.

NicEastvillage: "move" is singular because it assignments a single move to each state

Issue #190, opened 2023-06-07
LCGS syntax update proposal


NicEastvillage:
Future work from 8th semester includes updates to the LCGS syntax in order to make it more consistent and less tedious. This issue is for discussing how that new syntax would look like. Let's first state the design goals:

- Remove inconsistent notation: Some declarations start with a keyword while others start with a variable name or other symbols. Our syntax should try to minimize the number of exceptions, e.g. by having all declarations start with a keyword.
- Tedious duplicate code: If label or update expressions depend on the same complex expression, then the only
solution currently, is to copy the expression. This should be solved with a abstraction like functions or similar.
- (Extra) Only labels can be used in ATL: It would be nicer to express ATL queries if they could directly use the state variables in LCGS. This change to the ATL syntax would make the label declaration somewhat unnecessary unless we make it more versatile.

### Initial proposal

I propose we update the syntax as follows:

- Replace labels with `expr` declarations.
  - These are values derived from the current state.
  - They can be used in ATL queries like labels.
  - They can have parameters, making them similar to functions
- Add `dynexpr` declarations.
  - These are similar to expr declarations, but they are derived at transition-time, meaning they can use player actions in their definition.
- Add `var` keyword for variables, simply prepended to the current syntax
  - They `var` keyword is conviently 4 letters, which makes it align with an indented update expression.
- Add `action` keyword
  - Replaces the square brackets around actions names for more consistency
  - Action name and condition is instead separated by `:`
- Add explicit parameters to templates showing which identifiers must be relabeled
  - The characteristics of the parameters (e.g. whether it must be an identifier, or whether it must have certain fields) are infered for high versatility.
  - Whether relabeling is removed is TBD.

The Mexican Standoff would look like this with this proposed syntax:
```
const MAX_HEALTH = 2;

player billy = cowboy(clayton, jesse);
player clayton = cowboy(jesse, billy);
player jesse = cowboy(billy, clayton);

expr over = billy.dead + clayton.dead + jesse.dead >= 2;

template cowboy(target1, target2)
    
    var health : [0..MAX_HEALTH] init MAX_HEALTH;
        health' = max(health - target1.shoot_left - target2.shoot_right, 0);

    expr alive = health > 0;
    expr dead = !alive;
    expr something_with_params(a, b) = a * 2 > b;

    dynexpr hit = target1.shoot_left || target2.shoot_right;

    action wait : 1;
    action shoot_right : health > 0 && target1.health > 0;
[...]

Issue #42, opened 2020-11-24
[LCGS] Add compile errors


NicEastvillage:
Currently the program crashes with a sometimes-helpful message, whenever the LCGS program is faulty. Ideally it should give a (or multiple) compiler errors describing exactly what is wrong.

Additionally, the compile errors can be improved immensely if we calculate the spans of AST nodes. That is, the start and end position of the text that resulting in the AST node. `Span` is already defined in `lcgs::parse.rs` as well as a helper function to capture the spans of parsed text.

NicEastvillage:
I found a great article that will help us in this endeavour: https://eyalkalderon.com/blog/nom-error-recovery/

PR #216, opened 2023-09-07
Rewrite ATL parsing for better parse errors and more flexibility

NicEastvillage:
This PR rewrites the ATL parser. Instead of using `pom` it is now a handmade lexer and parser. This PR is the first step towards #42, #104, #190.

This PR includes the following:
- Replaces the ATL `pom` parser with handmade Lexer and Parser.
  - The Lexer and Parser is designed to be able to parse LCGS too in the future.
- Adds better parsing and semantic analysis errors. See below.
- Changes disjunction and conjunction in ATL from `&` and `|` to `&&` and `||`, respectively, to be consistent with LCGS.
- Removes JSON CSG/ATL query support (Its rarely used anyway).

Example of new error messages:
image


falkecarlsen (approved):
:shipit:

PR #220, opened 2023-09-12
Set workspace resolver to 2

NicEastvillage:
This gets rid of the following warning which has started to appear:
```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"
```


falkecarlsen (approved):
Can confirm fixes warning

PR #221, opened 2023-09-12
fmt main with new let-else rules

NicEastvillage:
Our main branch is not formatted correctly. The format rules for `let-else` statements have changed in the newer version of `rustfmt`.


NicEastvillage:
Resolved by #222

PR #222, opened 2023-09-12
CI-rework

falkecarlsen:
All CI tasks were run twice, and for some reason Clippy was run thrice. Now it still runs the same amount of times, but does so on our MSRV plus latest stable, additionally also runs a security audit.


NicEastvillage (approved):
LGTM

PR #226, opened 2023-09-26
Add new LCGS parser with better parse errors

NicEastvillage:
This PR adds a new LCGS parser. Instead of using `pom` it is now a handmade lexer and recursive-descent parser. This PR is the second step towards #42, #104, and #190.

This PR includes:
- AST nodes for LCGS programs
- Adds parsing of LCGS to new AST nodes
- Testing of new LCGS parser

This PR does NOT replace/remove the old LCGS parser. The old LCGS parser that uses `pom` is still in use for now as the semantic analysis of `IntermediateLCGS` (the symbol checking and relabeling) still uses the old AST. This will change in the next PR on this topic.


cgaal-engine/src/parsing/parser.rs
```diff
@@ -87,8 +90,337 @@ impl<'a> Parser<'a> {
...
+                    Expr::new_error()
+                )?;
+                let els = self.binary_expr(0)?;
+                let span = cond.span + els.span;
+                let kind = ExprKind::TernaryIf(cond.into(), then.into(), els.into());
+                Ok(Expr::new(span, kind))
+            }
+            _ => Ok(cond),
+        }
+    }
```
falkecarlsen, line 420: What if we have parenthesis in the sub-expr of a ternary?
```
a ? (x ? y : z) : c
```
This should be allowed.

cgaal-engine/src/parsing/parser.rs
```diff
@@ -178,6 +509,7 @@ impl<'a> Parser<'a> {
 
     /// Parse an expression term.
     pub fn term(&mut self) -> Result<Expr, RecoverMode> {
+        // TODO: min/max functions
```
falkecarlsen, line 512: Pushed back to later PR or forgotten?

falkecarlsen: Seems to be present in lines 575 584

falkecarlsen (commented):
Nice!

When migrating to new parser we should be aware if we are backwards-compat. If we are, then only minor-version needs to be bumped, otherwise another major bump required.

cgaal-engine/src/parsing/parser.rs
NicEastvillage: Yes. This is allowed already. Ternary expressions consist of binary expressions which consist of terms. Parenthesis is a term that allows ternary expressions inside, making your snippet ok.

cgaal-engine/src/parsing/parser.rs
NicEastvillage: I forgot to remove that. Good catch.

falkecarlsen (approved):
:shipit:

Issue #219, opened 2023-09-12
Add mixed/chaotic search strategy


NicEastvillage:
Our search strategies all have cases where they are better or worse. This means our user must know something about the shape of the dependency graph in order to pick the best search strategy for their case. However, our workers do not have to use the same search strategy. We should therefore add a search strategy builder that gives each worker a different search strategy, i.e. a mixed/chaotic search strategy (name TBD), and test its performance. Our expectation is that it would always perform averagely since it partially have the upsides and downsides of every search strategy.

In a similar vein, we could also allow the user to pick more than one search strategy, and give the workers strategies accordingly. E.g. `--search-strategy=bfs,bfs,dfs` would give two-thirds of the workers the BFS strategy and one-third the DFS strategy.

NicEastvillage:
Added in #233

PR #233, opened 2024-01-31
Add composite search strategies

NicEastvillage:
Adds the feature described in #219. The search strategy `-s mix` makes each worker use a different search strategy. The composite search strategy `-s bfs,bfs,rdfs,ihs` makes worker 1, 2, 5, 6, 9, 10, etc use BFS; worker 3, 7, 11 etc uses RDFS; and worker 4, 8, 12, etc use IHS.

Also fixes a bug where the user could not select search strategy from the cli.

Benchmarking is TODO.

### DIFF
--- a/cgaal-engine/src/algorithms/certain_zero/search_strategy/composite.rs
+++ b/cgaal-engine/src/algorithms/certain_zero/search_strategy/composite.rs
@@ -93,7 +93,7 @@ impl CompositeSearchStrategyBuilder {
         strategies: Vec<CompositeSearchStrategyOption>,
     ) -> CompositeSearchStrategyBuilder {
         assert!(
-            strategies.len() > 0,
+            !strategies.is_empty(),
             "Composite search strategy must have at least one component search strategy"
         );
         CompositeSearchStrategyBuilder {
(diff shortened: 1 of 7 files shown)


cgaal-engine/src/algorithms/certain_zero/search_strategy/composite.rs
```diff
@@ -0,0 +1,209 @@
...
+            strategies,
+            index: RefCell::new(0),
+        }
+    }
+
+    /// Create a new [CompositeSearchStrategyBuilder] employing each search strategy evenly.
+    pub fn one_of_each(game: IntermediateLcgs) -> CompositeSearchStrategyBuilder {
+        use CompositeSearchStrategyOption::*;
+        // General/best strategies first in case the user is using few workers
+        Self::new(game, vec![Dhs, Rdfs, Ihs, Bfs, Lrs, Dfs, Los, Lps])
```
falkecarlsen, line 110: Spicy but sane choice. We should be aware that this default is to be changed, if we discover speedups elsewhere.

falkecarlsen (approved):
LGTM :shipit:

